PR #78, opened 2022-05-20
cypress testing for pool creation

smartdevcom:
pool creation unit test


peetzweg:
Closed in favor of #628

Issue #93, opened 2022-05-25
Make use of new `SupportedAssets` Information


Pczek:
## Background

<img width="500"  src="https://user-images.githubusercontent.com/839848/170221759-031916e2-ce17-4c11-be05-85df98fe4086.gif">

We rely heavily on price oracles, this oracles need to be selected for each asset individually. 
As of now we are adding them manually for each chain and asset. This is available to the Fuse SDK and stored in the SDK
https://github.com/Midas-Protocol/monorepo/blob/5ee4847597990f998d6201d954d92729e68d7230/packages/sdk/src/Fuse/index.ts#L124

To make it more user friendly and less confusing we would like to show the supported assets to the Pool Admin to select. Similar to uniswap or sushi.

<img width="280" alt="Screenshot 2022-05-25 at 10 39 29" src="https://user-images.githubusercontent.com/839848/170220622-653b182d-ca96-4282-8507-31c996a44e8c.png">

<img width="280" alt="Screenshot 2022-05-25 at 10 39 13" src="https://user-images.githubusercontent.com/839848/170220633-76bc47dd-849a-4ac6-a0aa-4b5733920c98.png">

## Definition Of Done

The Add Asset Modal offers the user a list of the supported Assets instead of adding the address manually.

It offers the wrapped native asset at the top. This can be adapted in the ui code.
At some point we want to list FEI, TOUCH, TRIBE at the top probably.

## Tasks

- [ ] Create Modal to select Asset
- [ ] Show Wrapped Native token as quick select at the top as in uniswap/sushi

smartdevcom:
Done https://github.com/Midas-Protocol/monorepo/pull/110

smartdevcom:
Done https://github.com/Midas-Protocol/monorepo/pull/110

PR #110, opened 2022-05-26
removed deprecated isTruncated attribute

smartdevcom:
Now admin can easily add assets from the supported assets list

image

image


smartdevcom:
related to https://github.com/Midas-Protocol/monorepo/issues/93

PR #158, opened 2022-06-02
sdk optimizations

rhlsthrm:
## Description

SDK is fully reliant on having access to the contract artifacts, which defeats the purpose of having it OSS. Also, including all the built artifacts will make the SDK bundle size HUGE. Separating it involves:

- Removing any reliance on artifacts. Wherever they weren't needed, I removed them.
- In some cases, we still need bytecode, so I started hardcoding it.
- There are a couple of ABIs that we need which are not part of `deployments.json`

I tried to use https://github.com/ItsNickBarry/hardhat-abi-exporter to export the ABIs into a nice format. Unfortunately, this didn't work with our hardhat setup because we aren't using hardhat to build. In the meantime, I'll just hardcode the ones we need.

## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes


packages/sdk/src/types.ts
```diff
@@ -27,8 +27,10 @@ export type MinifiedOraclesContracts = MinifiedCompoundContracts;
 
 export interface InterestRateModel {
   init(
-    interestRateModelAddress: string,
+    iinterestRateModelAddress: string,
```
carlomazzaferro, line 30: typo?

carlomazzaferro:
I like this approach a lot, should already help a good amount with bundle size

PR #172, opened 2022-06-03
replaceing word 'unrecognized Oracle' to 'MasterPriceOracle'

smartdevcom:


peetzweg:
Hmmm feels wrong. I wonder why there is not data.oracle? 🤔

carlomazzaferro:
> Hmmm feels wrong. I wonder why there is not data.oracle? 🤔

As far as I understand it comes from here: https://github.com/Midas-Protocol/monorepo/blob/development/packages/sdk/src/Fuse/index.ts#L398 

which isn't working, as far as I can tell

But also, not sure if it matters, because we only use the MPO

peetzweg:
We are getting the oracle name here:
https://github.com/Midas-Protocol/monorepo/blob/2d07cf59d8fb4ebbe87b430c542d206a87db7a3e/packages/ui/hooks/fuse/useExtraPoolInfo.ts#L21

And the name is resolved in this function of the bytecode is the same.

https://github.com/Midas-Protocol/monorepo/blob/2d07cf59d8fb4ebbe87b430c542d206a87db7a3e/packages/sdk/src/Fuse/index.ts#L401-L409

Hardcoding "Master Price Oracle" is not solving this bug. We either remove all of this oracle name resolving code or we fix it properly IMHO.

carlomazzaferro:
> We are getting the oracle name here:
> 
> https://github.com/Midas-Protocol/monorepo/blob/2d07cf59d8fb4ebbe87b430c542d206a87db7a3e/packages/ui/hooks/fuse/useExtraPoolInfo.ts#L21
> 
> And the name is resolved in this function of the bytecode is the same.
> 
> https://github.com/Midas-Protocol/monorepo/blob/2d07cf59d8fb4ebbe87b430c542d206a87db7a3e/packages/sdk/src/Fuse/index.ts#L401-L409
> 
> Hardcoding "Master Price Oracle" is not solving this bug. We either remove all of this oracle name resolving code or we fix it properly IMHO.

Yes I agree. I think we just need to change how the name resolving is working. We don't need to match runtime bytecode hash, we can just match addresses, because we do have the oracle addresses in the sdk in `sdk.availableOracles`:

https://github.com/Midas-Protocol/monorepo/blob/development/packages/sdk/src/chainConfig/index.ts#L12-L14

packages/ui/hooks/fuse/useExtraPoolInfo.ts
```diff
@@ -8,17 +8,16 @@ export const useExtraPoolInfo = (comptrollerAddress: string) => {
   const { data } = useQuery(['ExtraPoolInfo', currentChain.id, comptrollerAddress], async () => {
     if (comptrollerAddress) {
       const comptroller = fuse.createComptroller(comptrollerAddress);
+      const oracle = fuse.getPriceOracle(await comptroller.callStatic.oracle());
```
peetzweg, line 11: I think we could have left it in the Promise array, so everything is fetched simultaneously.

peetzweg (approved):

peetzweg:
Left a comment, feels better now. But please fix the build/lint errors before merging, thanks. 👍

carlomazzaferro (approved):

PR #175, opened 2022-06-03
Refactor `useFusePoolData` Fiat instead of Native

peetzweg:
We overwrote the `Native` price with the USD price. Super confusing. Now added fields for Fiat prices alongside the Native Price.

Important Part:

https://github.com/Midas-Protocol/monorepo/blob/51c6ac7808121d57700d15afdedc9312646838f2/packages/ui/hooks/useFusePoolData.ts#L7-L22

Still not the best way to do this. It's just a bit less confusing now. Next refactor comes in with new backend api probably.


smartdevcom (approved):

carlomazzaferro (approved):
confusing indeed, but appreciate the rework

PR #189, opened 2022-06-08
Pause mint market task

vminkov:
## Description

Hardhat task to pause the minting on a market. It also sets the pause guardian to the admin address, if it is not set yet


packages/sdk/tasks/pauseMarketMinting.ts
```diff
@@ -0,0 +1,43 @@
+import { task, types } from "hardhat/config";
+import { CToken } from "../lib/contracts/typechain/CToken";
+import { Comptroller } from "../lib/contracts/typechain/Comptroller";
```
carlomazzaferro, line 3: these imports are unlikely to work -- in my experience, you can't do such imports in tasks. Did you test this locally?

vminkov: this worked, right?

PR #193, opened 2022-06-09
fuse should be initialized after signer is updated

smartdevcom:
Now `fuse` will be updated after signer is changed successfully, we can still see `underlying network changed` error in browser console (seems to be related to `wagmi`) but `fuse` is updated with correct signer so we can get all pools successfully


packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList.tsx
```diff
@@ -65,12 +69,16 @@ const FusePoolList = () => {
...
+    setErr(error as Err);
+  }, [error]);
+
+  if (err && err.code !== 'NETWORK_ERROR') {
     return (
       <AlertHero
         status="warning"
         variant="subtle"
-        title="Unexpected Error"
+        title={err.reason ? err.reason : 'Unexpected Error'}
```
peetzweg, line 81: why not use `error` directly? It should cause a rerender the same way as setting a state does. 🤔

peetzweg (approved):
If this helps, we should revaluate it on the development deployment. 🤞

peetzweg (changes requested):
I can't switch chains via metamask anymore.

peetzweg:
On the first load I can't change the chain properly via metamask.

Also the chain id is not always included or correct in the URL.

It should always state the correct chainId in the URL.
See in demo, the first chain switch to bsc from evmos is not working, it gets stuck and than I'm even asked to get back to evmos.
Second time it works. But than the chainId in the URL is not correctly changed. It is always 9000 instead of chapel or moonbases chain id.

Screenshot 2022-06-08 at 17 45 17

peetzweg (approved):
Just tested it again, it works quite well and better than the current implementation. lets take this upgrade from here for now :)

@smartcondev can you fix the gh action error and conflict. Than it's ready to be merged. 👍

peetzweg:
Leaving this here for future reference:
https://discord.com/channels/930077813585899561/930157130454024192/98481724735055875

Screenshot 2022-06-13 at 14 20 46
0

PR #196, opened 2022-06-10
build and deploy netlify on ci

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/.github/workflows/package-ui-test.yaml
+++ b/.github/workflows/package-ui-test.yaml
@@ -1,4 +1,4 @@
-name: UI/PR - build, lint and test
+name: UI/PR - build, lint, test & deploy UI
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/package-ui-test.yaml
```diff
@@ -47,3 +47,44 @@ jobs:
...
+      - name: Install Netlify CLI
+        run: npm intall -g netlify-cli
+
+      - name: Link with project
+        run: |
+          netlify link --id ${{ secrets.NETLIFY_SITE_ID }}
+
+      - name: Deploy with Netlify
+        if: github.ref == 'refs/heads/development'
+        run: netlify deploy --branch development --prod --build
```
peetzweg, line 86: We have netlify setup to automatically trigger a build once we push to main/development, we don't need to start it manually here.


Furthermore we could make this `deploy` job dependent on the `build` job and just trigger a new deployment via the netlify cli if you want?

peetzweg: Ah nevermind, will adapt and deactivate auto deployment via netlify.

carlomazzaferro: > We have netlify setup to automatically trigger a build once we push to main/development, we don't need to start it manually here.

The whole point is doing the build here, so we can use the whole forge toolchain and get rid of the `out` directory files, and build the contracts on the fly

carlomazzaferro: Basically, what I want is be able to build + deploy via GHA only :) if that makes sense

peetzweg: yep yep got it. Was a bit confused. Makes sense, wasn't aware the `--build` flag causes the cli to call the build command specified in the netlify.toml.

peetzweg: The build step in the toml however, would miss the command to generate the `out` folder themselves, probably `forge install` `forge compile` or so.

peetzweg: Just checked to deactivate the automatic deployment on netlify. Even when set to `none` it will trigger deployment of the production branch . xD
Maybe we just need to setup a branch to production whcih we don't use. Something like a `donotuse` branch or so.

Screenshot 2022-06-10 at 10 04 08

packages/sdk/package.json
```diff
@@ -16,8 +16,8 @@
     "deployments.json"
   ],
   "scripts": {
-    "build": "npm run build:forge; npm run typechain && npm run build:sdk",
-    "build:forge": "cd lib/contracts && forge build",
+    "build": "npm run build:forge && npm run typechain && npm run build:sdk",
```
carlomazzaferro, line 19: guess you've resolved your prev question

PR #198, opened 2022-06-10
Prevent Deploy Step on Pull Requests

peetzweg:
## Description

Before this PR, the `deploy` job was always run, so the build was almost fully executed just to stop at the end to trigger the actual deployment build.

I've split up the original GH action into two files, one for PRs to `development`, `main`, which just does the test and lint stuff.
And the other one only for `push` to actually build and trigger netlify deployment.


carlomazzaferro:
Great call!

carlomazzaferro (approved):

PR #199, opened 2022-06-10
SDK test updated for utils

dezine2dev:
## Description

SDK sample unit test for Fuse/utils


packages/sdk/tests/new/Fuse/utils.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  let keccak256Spy: SinonSpy;
+  let getCreate2AddressSpy: SinonSpy;
+  let ContractFactoryStub: SinonStub;
+  let ContractStub: SinonStub;
+  let AbiCoderStub: SinonStub;
+
+  let utilsFns;
+
+  beforeEach(() => {
+    utilsFns = proxyquire("../../../src/Fuse/utils", {
```
dezine2dev, line 18: I am using proxyquire here to proxy nested function. That was not possible to do with sinon.
Let me know if this is not ok

rhlsthrm: https://github.com/Midas-Protocol/monorepo/blob/sdk-test/packages/sdk/tests/new/Fuse/index.spec.ts#L6

Did you see how this was done?

rhlsthrm (changes requested):

peetzweg:
Anything holding this back?

Issue #184, opened 2022-06-07
Calculate Pools Utilisation Only Based On Borrowable Assets


peetzweg:
## Background

The current utilisation value of a pool is calculated based on all asset in a pool. However, this is not always correct as some assets are not able to be borrowed.

The utilisation calculation should only take markets into account which are available to be borrowed.

<img width="705" alt="Screenshot 2022-06-07 at 15 36 05" src="https://user-images.githubusercontent.com/839848/172395266-fae7121a-d980-47d7-a9f4-de016150c2b9.png">

Maybe we can move this calculation in the SDK?
#12

PR #201, opened 2022-06-13
moved utilization calculation to the SDK

smartdevcom:


packages/ui/hooks/useFusePoolData.ts
```diff
@@ -32,6 +32,7 @@ export const useFusePoolData = (poolId: string) => {
 
       const res = await fuse.fetchFusePoolData(poolId, address);
       const assetsWithPrice: MarketData[] = [];
+      console.log(res.assets);
```
peetzweg, line 35:
```suggestion

```

packages/ui/components/pages/Fuse/FusePoolPage/index.tsx
```diff
@@ -112,14 +112,7 @@ const FusePoolPage = memo(() => {
...
                   label="Utilization"
-                  value={
-                    data
-                      ? data.totalSuppliedFiat.toString() === '0'
-                        ? '0%'
-                        : ((data?.totalBorrowedFiat / data?.totalSuppliedFiat) * 100).toFixed(2) +
-                          '%'
-                      : undefined
-                  }
+                  value={data ? data.utilization.toFixed(2) + '%' : undefined}
```
peetzweg, line 115: this pr does not seem to contain the code which is setting the `data.utilization` value. Was it already available all along?

packages/sdk/src/modules/FusePools.ts
```diff
@@ -96,13 +99,25 @@ export function withFusePools<TBase extends FuseBaseConstructor>(Base: TBase) {
...
 
         asset.liquidityNative =
           Number(utils.formatUnits(asset.liquidity)) * Number(utils.formatUnits(asset.underlyingPrice));
 
         totalLiquidityNative += asset.liquidityNative;
+
+        if (!asset.isSupplyPaused) suppliedForUtilization += asset.totalSupplyNative;
+        if (!asset.isBorrowPaused) borrowedForUtilization += asset.totalBorrowNative;
+        if (suppliedForUtilization !== 0) {
+          utilization = (borrowedForUtilization / suppliedForUtilization) * 100;
```
peetzweg, line 119: This calculation can be done outside of the for loop.

peetzweg: But maybe this PR is more about moving hte calculation outside the FE and into SDK and not about this ticket here yet, right?

https://github.com/Midas-Protocol/monorepo/issues/184

smartdevcom: Yes but also this PR fixed `utilization` calculation based on only borrow-able assets

smartdevcom: makes sense

smartdevcom: > Also not sure if this calculation is correct. thinking
> 
> Shouldn't it be something like `currently borrowed / total amount which is abled to borrowed` ?
> 
> Where as `total amount which is abled to borrowed` is `sum of all the supply of borrowable assets`, so not including the supply of assets which are not borrowable.
> 
> `currently borrowed` => `sum of all borrowed assets`

Check this https://github.com/Midas-Protocol/monorepo/pull/201/commits/a965ecbd72e6a7a9c5f866f261e5004f600463c5

peetzweg: Perfect, I think this should work as intended in #184

peetzweg (approved):

Issue #12, opened 2022-05-13
RESEARCH: Move Important Calculations from UI to SDK


peetzweg:
We have several calculations which would be better suited in the SDK Code.
We should research which calculations these are and bring them over into the SDK.

These new functions should be unit tested.

Unqualified suggestions
+ Borrow calculations
+ supply calculations
+ repay calculations
+ apy stuff?
+ borrow limits

smartdevcom:
> useUpdatedUserAssets

Done with this PR https://github.com/Midas-Protocol/monorepo/pull/218

smartdevcom:
> The `useUpdatedUserAssets` hook is a great candidate to add to the SDK itself to benefit from sharing this logic and test coverage.
> 
> https://github.com/Midas-Protocol/monorepo/blob/8d7dee9f772cc4462b20efb2f0ca6b1913af1c4b/packages/ui/hooks/fuse/useUpdatedUserAssets.ts#L10-L20

Done with this PR https://github.com/Midas-Protocol/monorepo/pull/218

peetzweg:
Great 👍 
Let's close this for now and just open new tickets if we find more good candidates to pass over to the SDK.

PR #218, opened 2022-06-15
moved useUpdatedUserAssets hook from UI to the SDK

smartdevcom:
Moved `useUpdatedUserAssets` hook from UI to SDK. related to https://github.com/Midas-Protocol/monorepo/issues/12


peetzweg:
👍 Let's add some tests for this as it now lives in the SDK package. 🙌

PR #224, opened 2022-06-16
fixed APR, APY calculation

smartdevcom:


packages/ui/hooks/fuse/usePoolDetails.ts
```diff
@@ -14,14 +17,28 @@ export const usePoolDetails = (assets: NativePricedFuseAsset[] | undefined) => {
...
-          convertMantissaToAPY(topLendingAPYAsset.supplyRatePerBlock, 365)
+          convertMantissaToAPY(
+            asset.supplyRatePerBlock,
+            getBlockTimePerMinuteByChainId(currentChain.id),
+            365
+          ) >
+          convertMantissaToAPY(
+            topLendingAPYAsset.supplyRatePerBlock,
+            getBlockTimePerMinuteByChainId(currentChain.id),
+            365
```
peetzweg, line 28: we could maybe make this optional and the default value 356?

peetzweg:
Hm how about we move these functions also to the SDK, @smartdevcom? 
We maybe even could calculate these values before and just put them also in the `PoolData`? And also add some happy path tests for some values. To just make sure they are correct.

`convertMantissaToAPY`
`convertMantissaToAPR`


I think this could be helpful.

packages/ui/hooks/fuse/usePoolDetails.ts
smartdevcom: > 



> Hm how about we move these functions also to the SDK, @smartdevcom? We maybe even could calculate these values before and just put them also in the `PoolData`? And also add some happy path tests for some values. To just make sure they are correct.
> 
> `convertMantissaToAPY` `convertMantissaToAPR`
> 
> I think this could be helpful.

Sure, I will do it

peetzweg:
@smartdevcom can we merge this and do the move into sdk later? I think I need this sooner than i thought :D

PR #270, opened 2022-06-22
user can type percentage

smartdevcom:
image


peetzweg:
Slider and arrows work, but typing is working weird. It's inserting other numbers than expected or loses focus at all.

demo

peetzweg:
Maybe something like `react-hook-form` could help us build better forms. I have used this and `formik` in the past with great success. 
Could be a great first try to use it at midas.

Both seem to work with chakra ui

https://chakra-ui.com/getting-started/with-hook-form

smartdevcom:
image

`react-hook-form` is integrated to our UI and it worked very well. And also loosing focus when typing in input field is fixed.

peetzweg:
Wow this is soooo much nicer than before! Great use. Haven't checked the code in detail but it works super well. 👍 
We should use it in more In
put cases in our app! :)


Screenshot 2022-06-23 at 13 49 01
Screenshot 2022-06-23 at 13 48 55

peetzweg (approved):

PR #280, opened 2022-06-24
borrow limit slider on borrow modal

smartdevcom:
image

borrow amount input and slider are synced now. And updated slider showcases.


peetzweg:
Wow this is so cool. Works great! 👏 

Had this issue on production though. Can you reproduce it? It shows two sliders even when nothing is borrowed on BSC.

Screenshot 2022-06-24 at 12 08 09


Seem to be handled correctly on local

Screenshot 2022-06-24 at 12 14 12

smartdevcom:
On my side, it looks like this
image

smartdevcom:
works now, this issue occurred because so tiny amount is borrowed already so couldn't calculate it's width on slider
image

Issue #255, opened 2022-06-21
Borrow Modal shows Changed APR although Amount is Zero


peetzweg:
## Background

In the production app, in the borrow modal the APR is updated although the amount to borrow is `0`. So from my understanding the APR should not be shown as changing at all.

<img width="569" alt="Screenshot 2022-06-21 at 12 04 27" src="https://user-images.githubusercontent.com/839848/174774400-c376f6a8-b4c6-4a58-b0a5-aec0f5aed261.png">

## Definition Of Done

We know why this is updating although we don't want to borrow anything or it is not updating when not borrowing anything

peetzweg:
@smartdevcom Can you check this. It's weird that the value changes even if the input is "zero".
I remember you moved it to the sdk, do we have tests for it as well?

peetzweg:
> This seems to be related to `updatedAsset.borrowRatePerBlock` calculation. Because `asset.borrowRatePerBlock` is different from `updatedAsset.borrowRatePerBlock` even borrow input is `0`
> 
> https://github.com/Midas-Protocol/monorepo/blob/a00baf8d14f5b94e740d43c1aa67013de7faf33b/packages/sdk/src/modules/Asset.ts#L250-L253

Hmm, weird are you sure? `borrowRatePerBlock` is essentially the APR the user has to pay for borrowing. It should be the same if the "amount" which is added is zero.

@smartdevcom can you add some more sophisticated tests to this suite right here.
Especially the amount zero cases should be handled properly, basically, no value should change afaik.

https://github.com/Midas-Protocol/monorepo/blob/45a74d01ab44066dc4562aee457896ad11656c85/packages/sdk/tests/unit/AssetModule.spec.ts#L13-L83

smartdevcom:
@peetzweg , Could you check this again?

peetzweg:
Seems to work fine again 👍

PR #319, opened 2022-07-01
test for getting difference between asset.borrowRatePerBlock and one …

smartdevcom:
While I am trying to debug this issue https://github.com/Midas-Protocol/monorepo/issues/255, I found that `borrowRatePerBlock`s from asset's property and from irm were different.

See my logs 

image

`borrowRatePerBlockByIrm` should be the same as `asset.borrowRatePerBlock` but different


smartdevcom:
This issue is Done

PR #339, opened 2022-07-06
`FuseFlywheelDynamicRewards_` => `FuseFlywheelDynamicRewardsPlugin_`

peetzweg:
update name to make transparent which contract is actually used


packages/sdk/deployments/bsc/DotDotLpERC4626_0xf0a2852958aD041a9Fb35c312605482Ca3Ec17ba.json
```diff
@@ -0,0 +1,1234 @@
+{
+  "address": "0x3c29e9b0CfE6FfF97f373eAbEADE9475FaC3bd4e",
```
vminkov, line 2: why does the name of this file contain an address suffix?

packages/sdk/tasks/jarvisFix.ts
```diff
@@ -0,0 +1,92 @@
...
+      "0x1B6E11c5DB9B15DE87714eA9934a6c52371CfEA9", //_asset,
+      dddFlywheelAddress, //_dddFlywheel,
+      epxFlywheelAddress, //_epxFlywheel,
+      "0x8189F0afdBf8fE6a9e13c69bA35528ac6abeB1af", //_lpDepositor,
+      marketAddress, //_rewardsDestination,
+      ["0x84c97300a190676a19D1E13115629A11f8482Bd1", "0xaf41054c1487b0e5e2b9250c0332ecbce6ce9d71"], //_rewardTokens
+    ];
+    console.log({ strategyArgs });
+
+    const pluginDeployment = await hre.deployments.deploy("DotDotLpERC4626_" + marketAddress, {
```
vminkov, line 43: is this the new convention or just a one-time thing? @carlomazzaferro

vminkov (commented):
probably the .json files with the new plugins deployments will become unnecessary, too?

carlomazzaferro:
> probably the .json files with the new plugins deployments will become unnecessary, too? 

That's right

packages/sdk/tasks/jarvisFix.ts
carlomazzaferro: That's the convention. As we've discussed, each market will have its own plugin address

packages/sdk/deployments/bsc/DotDotLpERC4626_0xf0a2852958aD041a9Fb35c312605482Ca3Ec17ba.json
peetzweg: It's the address of the market it was deployed for.

Issue #259, opened 2022-06-21
Issue withdrawing the 2brl LP


NikMidas:
Darrell came across an issue with withdrawing the max amount in 2BRL LP.

![Image](https://user-images.githubusercontent.com/103433798/174607179-04bf1048-3d69-4047-a90d-0d055fe188f8.png)

TXN: https://bscscan.com/tx/0x4df2d43f200a109398e6482b9c698ec8538c6656c75c3aa6625d063153f716fb

smartdevcom:
![image](https://user-images.githubusercontent.com/45715420/177959730-f4732314-21c7-45aa-a2ee-63c1753e3472.png)

I can withdraw max value now

NikMidas:
We have another issue open regarding this item.  Closing this in favor of the other one.

peetzweg:
https://github.com/Midas-Protocol/monorepo/issues/431

PR #351, opened 2022-07-08
fix issue with maxAmount calculation

dezine2dev:
## Description

Fix max withdraw issue and other potential rounding issue

## Type of change

- [ ] Refactoring
- [x] Bug fix (non-breaking change which fixes an issue)

The issue comes with managing max Amount. Current logic is converting fetched maxAmount big Number to string value and change that to big Number again on Confirm click. At this point there is tiny difference by rounding issue.
For max button case, we removed coverting process and use fetched maxAmount directly on final Confirm.

## Related Issue(s)

Fixes [the issue](https://github.com/Midas-Protocol/monorepo/issues/259)

### DIFF
--- a/packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
+++ b/packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
@@ -130,9 +130,9 @@ const AmountSelect = ({
   const updateMaxBNAmount = (bNAmount: BigNumber) => {
     const str = utils.formatUnits(bNAmount, asset.underlyingDecimals);
     _setUserEnteredAmount(str);
-    _setAmount(bNAmount)
+    _setAmount(bNAmount);
     setUserAction(UserAction.NO_ACTION);
-  }
+  };
 
   const { data: amountIsValid } = useQuery(['ValidAmount', mode, amount], async () => {
     if (amount === null || amount.isZero()) {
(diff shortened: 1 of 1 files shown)


carlomazzaferro (approved):

carlomazzaferro:
LG, except failing CI

smartdevcom:
I think it won't change even after converting twice. we use `formatUnits` and `parseUnits`. @dezine2dev  could you reproduce this kind of issue?
Like I tried `console.log(utils.parseUnits(utils.formatUnits('0x91929395471863899293')));` returns `0x91929395471863899293`

dezine2dev:
> I think it won't change even after converting twice. we use `formatUnits` and `parseUnits`. @dezine2dev could you reproduce this kind of issue? Like I tried `console.log(utils.parseUnits(utils.formatUnits('0x91929395471863899293')));` returns `0x91929395471863899293`

Yes you are right. Double converting is not an issue but updateAmount function has a problem for handling maxAmount. 
```
 const bigAmount = utils.parseUnits(
      toFixedNoRound(Number(newAmount), tokenData?.decimals || DEFAULT_DECIMALS),
      tokenData?.decimals
    );
```
We are rounding fetched Maxamount at this line and it does not let us completely wipe out the max amount. I also see tiny more issues by this.
for example
- sometimes confirm button becomes disable after hitting max button due to max amount exceeding (from tiny difference)
- max repay has also tiny difference issue.

smartdevcom:
We should round down the fetched amount because sometimes it exceeds 18 numbers under decimal point otherwise we get an error with `utils.parseUnits` and also the value which is rounded down should be smaller than real max amount so confirm button should not be disabled if max amount is correct.
@dezine2dev I am not sure which issue you faced but could you post a screenshot or video or whatever so I can fix that

smartdevcom:
Done here https://github.com/Midas-Protocol/monorepo/pull/411

PR #359, opened 2022-07-13
set configuration for polygon mainnet and testnet

smartdevcom:
Added polygon mainnet and testnet configuration


carlomazzaferro:
https://github.com/Midas-Protocol/monorepo/pull/358

PR #398, opened 2022-07-26
390 show apy from backend in for plugins and flywheels

smartdevcom:
apy is implemented for asset with plugin + dynamic flywheel and with only plugin.
In jarvis pool, for 2brl asset, api is calling `http://localhost:3000/api/apyData?underlyingAddress=0x1B6E11c5DB9B15DE87714eA9934a6c52371CfEA9&&pluginAddress=0x3c29e9b0CfE6FfF97f373eAbEADE9475FaC3bd4e&&rewardAddress=0xAf41054C1487b0e5E2B9250C0332eCBCe6CE9d71` but pluginAddress doesn't matched in Supabase data so apy doesn't show up here

image

image

And for an asset with only plugin (without dynamic flywheel), it shows `+ <plugin icon> <apy>`

image

Also tooltip shows `No Plugin` word because plugin address is mismatched or isn't upgraded in `chainPluginConfig` in the SDK

image


smartdevcom:
image

It seems that `fuse.chainPlugins[marketAddress]` was removed but we need this when deploying asset to show plugins user can choose cc @peetzweg

peetzweg:
> image
> 
> It seems that `fuse.chainPlugins[marketAddress]` was removed but we need this when deploying asset to show plugins user can choose cc @peetzweg

Yes thats correct. I will probably remove this stuff as well. 
We currently don't allow users to create pools anymore and it's also not possible for them to add plugins directly themselves. They need to deploy their own plugin. We need to refactor this a bit more to find out how this can be properly done.

PR #435, opened 2022-08-03
updated polygon plugin asset addresses

coolnft:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


peetzweg:
Looks good to me. I just wonder if we maybe need to make the names of the plugins more specific as they are probably used only by a single market. So if we have another pool with i.e. `AGEURJEUR` it will collide with the already deployed strategies.

peetzweg:
Can you delete the old json files in the deployments folder without the market address suffix. Than we are good to go :)

peetzweg (approved):

PR #443, opened 2022-08-04
One Time Tasks for Setting Up Jarvis Pool on Polygon

peetzweg:
## Description
Adds these 4 new jarvis specific task which should be run in the following order:

`jarvis:polygon:deploy-plugins` (coolnft already deployed the plugins in a previous pr, so we are using these addresses in the following tasks)

`jarvis:polygon:whitelist-plugins`

`jarvis:polygon:upgrade-implementations`

`jarvis:polygon:set-plugins`


packages/sdk/src/chainConfig/assets/polygon.ts
```diff
@@ -218,28 +218,28 @@ export const assets: SupportedAsset[] = [
   },
   {
     symbol: assetSymbols["JEUR-PAR"],
-    underlying: "0x0f110c55efe62c16d553a3d3464b77e1853d0e97",
+    underlying: "0x0f110c55EfE62c16D553A3d3464B77e1853d0e97",
```
carlomazzaferro, line 221: are these case changes intentional?

peetzweg: Yes, usually the ui works with the addresses direclty from the contract and never lowercases them. So we should use the correctly formatted address here.

packages/sdk/tasks/one-time/jarvis-polygon-plugins.ts
```diff
@@ -0,0 +1,104 @@
+import { task, types } from "hardhat/config";
+
+const POOL_NAME = "Jarvis jFIAT";
+const COMPTROLLER = "0xD265ff7e5487E9DD556a4BB900ccA6D087Eb3AD2";
+
+const UNDERLYINGS = {
+  "agEUR-jEUR": "0x2ffbce9099cbed86984286a54e5932414af4b717",
```
carlomazzaferro, line 7: Don't particularly love having the underlyings here as we already have them all in `assets/polygon.ts`, as well as the symbols, but for now that's alright I guess

packages/sdk/tasks/market/set-plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+    const midasSdkModule = await import("../../tests/utils/midasSdk");
+    const sdk = await midasSdkModule.getOrCreateMidas();
+
+    const pool = await poolModule.getPoolByName(poolName, sdk);
+
+    const assets = pool.assets;
+
+    const assetConfig = assets.find((a) => a.underlyingToken === underlying || a.underlyingSymbol === underlying);
+
+    const market = pool.assets.find((a) => a.underlyingToken == assetConfig.underlyingToken);
```
vminkov, line 27: querying twice `pool.assets` seems illogical here

packages/sdk/tasks/market/set-plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+    console.log("market", market);
+
+    const cTokenInstance = sdk.getCErc20PluginInstance(market.cToken);
+
+    console.log(`Setting plugin to ${pluginAddress}`);
+    const setPluginTx = await cTokenInstance._updatePlugin(pluginAddress);
```
vminkov, line 33: check if the deployer has admin rights

packages/sdk/tasks/one-time/jarvis-bsc-fix.ts
```diff
@@ -0,0 +1,140 @@
+/**
```
vminkov, line 1: these one-time files are going to be removed after the deploy or refactored for general usage?

peetzweg: probably just removed after use. Maybe kept for reference. It just now the most feasible way to get the pool up and understanding what the process is to put it into place in the frontend somehow.

packages/sdk/tasks/market/set-plugin.ts
peetzweg: I guess it will just through an ANAUTHORIZED exception or similar right? 

But true it would be good to better implement the more commonly used tasks.

packages/sdk/tasks/market/set-plugin.ts
peetzweg: copied it from upgrade market task. I guess it's needed here to be able to allow for SYMBOL to put in. I prefer getting the actual address better also.

packages/sdk/tasks/one-time/jarvis-polygon-plugins.ts
peetzweg: 100% 
true true, maybe good to import them from polygon. But yes it's a one time shot. So maybe not worth the effort if they are at least the same :D

Issue #422, opened 2022-08-01
Deploy Beefy Finance plugin on Polygon, and add support for 2EUR, 2JPY, 2CAD, 2SGD assets


carlomazzaferro:
Beefy strategies:

- https://app.beefy.com/vault/jarvis-2eur
- https://app.beefy.com/vault/jarvis-2eurp
- https://app.beefy.com/vault/jarvis-2jpy2
- https://app.beefy.com/vault/jarvis-2cad
- https://app.beefy.com/vault/jarvis-2sgd

They all use the BeefyVaultV6, for which we already have support for

carlomazzaferro:
This will require a custom hardhat task to deploy a single plugin for each of the markets. Akin to: https://github.com/Midas-Protocol/monorepo/blob/8b49320bc1208b16b8ac83a5d774908bf77b60f4/packages/sdk/tasks/jarvisFix.ts#L6-L91 

But without the dynamic rewards toolchain

peetzweg:
For reference

https://app.beefy.com/vault/jarvis-2eur `0x2ffbce9099cbed86984286a54e5932414af4b717`
https://app.beefy.com/vault/jarvis-2eurp `0x0f110c55EfE62c16D553A3d3464B77e1853d0e97`
https://app.beefy.com/vault/jarvis-2jpy2 `0xaA91CDD7abb47F821Cf07a2d38Cc8668DEAf1bdc`
https://app.beefy.com/vault/jarvis-2cad `0xA69b0D5c0C401BBA2d5162138613B5E38584F63F`
https://app.beefy.com/vault/jarvis-2sgd `0xeF75E9C7097842AcC5D0869E1dB4e5fDdf4BFDDA`

peetzweg:
New assets we should add:

`2eur (EURT) pool: https://polygonscan.com/address/0x2c3cc8e698890271c8141be9f6fd6243d56b39f1`

`actually 2nzd is missing also: https://polygonscan.com/address/0x976a750168801f58e8aedbcff9328138d544cc09`

`add jCHF, jGBP, jCNY, jMXN, etc. in the assets you can borrow or lend`

PR #452, opened 2022-08-05
Feat/more assets for jarvis

peetzweg:
Adds functionality to setup the requested markets:

https://github.com/Midas-Protocol/monorepo/issues/422#issuecomment-1206177350


peetzweg:
@carlomazzaferro our `market:create` task is quite limited and it requires hardcoded `assetConfig`. We should refactor it in a more general version at some point.

Not sure about the parameters for these markets. Can you deploy the markets with the correct settings. 🤷 
I can than deploy the plugins and upgrade the markets. ✌️

PR #454, opened 2022-08-05
add assets that are missing, fix docs and oracles

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


coolnft (approved):
changes looks good to me.

packages/sdk/src/chainConfig/assets/polygon.ts
```diff
@@ -221,40 +221,54 @@ export const assets: SupportedAsset[] = [
     underlying: "0x2ffbce9099cbed86984286a54e5932414af4b717",
```
peetzweg, line 221: This one is using the checksum address so mixed case version. Every other address seems to be just lowercased.

carlomazzaferro: changed to `0x2fFbCE9099cBed86984286A54e5932414aF4B717` -- is this correct?

carlomazzaferro: looks like it :)

PR #463, opened 2022-08-08
updates contract

smartdevcom:
This will sort assets by its `underlyingSymbol`. But assets list shows by their `cToken.symbol` and it is different from `underlyingSymbol` for some assets.

image

So we need to add some extra info (like cToken data) into `assets` data to sort them. wdyt? cc @peetzweg


packages/ui/utils/sortAssets.ts
```diff
@@ -0,0 +1,15 @@
+import { SupportedAsset } from '@midas-capital/types';
+
+import { MarketData } from '@ui/hooks/useFusePoolData';
+
+export const sortAssets = (assets: MarketData[]) => {
+  return assets.sort((a, b) => {
+    return b.underlyingSymbol.toLowerCase() > a.underlyingSymbol.toLowerCase() ? -1 : 1;
```
peetzweg, line 7: does this work?
We may better use `localeCompare` as often recommened for string sorting, afaik.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

peetzweg (approved):

PR #468, opened 2022-08-08
refactor for more logical structure

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/sdk/chainDeploy/mainnets/polygon.ts
+++ b/packages/sdk/chainDeploy/mainnets/polygon.ts
@@ -8,11 +8,18 @@ import {
   ChainlinkFeedBaseCurrency,
   deployChainlinkOracle,
   deployCurveLpOracle,
+  deployJarvisSynthereumLiquidator,
   deployUniswapLpOracle,
   deployUniswapOracle,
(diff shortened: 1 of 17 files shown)


packages/sdk/chainDeploy/mainnets/polygon.ts
```diff
@@ -437,6 +449,41 @@ const gelatoAssets: GelatoGUniAsset[] = [
   },
 ];
 
+const jarvisLiquidityPools: JarvisLiquidityPool[] = [
+  // jAUD -> USDC
+  { expirationTime: 40 * 60, liquidityPoolAddress: "0x36572797Cc569A74731E0738Ef56e3b8ce3F309c" },
```
vminkov, line 454: We have to ask the Jarvis team if we can use 0 for `expriationTime` , because in our use case it is used for a single atomic transaction (the liquidation) and we don't have the ability to wait for a later time for the mint/redeem to happen.

vminkov (approved):
it's good, let's merge it and we can then later ask the Jarvis team about the expiration time

PR #491, opened 2022-08-11
fixed hardcoded

smartdevcom:
I checked UI package for hardcoded and it seems to be all good

### DIFF
--- a/packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetSettings.tsx
+++ b/packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetSettings.tsx
@@ -75,16 +75,16 @@ export const AddAssetSettings = ({
     formState: { errors },
   } = useForm({
     defaultValues: {
-      collateralFactor: 50,
-      reserveFactor: 10,
-      adminFee: 5,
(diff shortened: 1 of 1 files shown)


packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetSettings.tsx
```diff
@@ -75,16 +75,16 @@ export const AddAssetSettings = ({
     formState: { errors },
   } = useForm({
     defaultValues: {
-      collateralFactor: 50,
-      reserveFactor: 10,
-      adminFee: 5,
+      collateralFactor: COLLATERAL_FACTOR.DEFAULT,
```
peetzweg, line 78: Where are these values stored? Were these already set somewhere o.O?

peetzweg (approved):

packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetSettings.tsx
smartdevcom: This is initial value and stored here https://github.com/Midas-Protocol/monorepo/blob/1a8ee242857f21ba24ef080e2bd97c81b9d614c6/packages/ui/constants/index.ts#L23

PR #503, opened 2022-08-15
Fix/refactor network data

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/chains/src/bsc/assets.ts
```diff
@@ -7,6 +7,7 @@ export const assets: SupportedAsset[] = [
     name: "Wrapped Binance Network Token",
     decimals: 18,
     oracle: OracleTypes.FixedNativePriceOracle,
+    extraDocs: "",
```
carlomazzaferro, line 10: Will be slowly adding these over time @peetzweg

smartdevcom (approved):

peetzweg:
👍 
Epic.

Maybe a new ticket though:


However, for ui we need to make sure can import individual packages for sure. Otherwise all the pages will just blow up with a 2.2MB chains packages.

We need to create multiple entry points to export more granular what we need.
https://nodejs.org/api/packages.html#package-entry-points

I tried that but wasn't able to finish it yet properly. It work in development but once building the ui it broke.

VSCode extension I've used:
https://marketplace.visualstudio.com/items?itemName=wix.vscode-import-cost

peetzweg (approved):

PR #504, opened 2022-08-16
Feature/document tokens

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/chains/src/bsc/assets.ts
```diff
@@ -1,258 +1,329 @@
...
   },
   // Jarvis
   {
     symbol: assetSymbols.jBRL,
-    underlying: "0x316622977073BBC3dF32E7d2A9B3c77596a0a603",
+    underlying: jBRL,
     name: "Jarvis Synthetic Brazilian Real",
     decimals: 18,
     oracle: OracleTypes.ChainlinkPriceOracleV2,
+    extraDocs: `You can acquire this asset on the <a href="https://v1-app.jarvis.exchange/" target="_blank" style="color: #BCAC83; cursor="pointer">Jarvis Network</a> website`,
```
peetzweg, line 230: We should leave out the `color` here and just style it via global style in the for `<a/>` or even better local styling of `<a/>` children of that tooltip.

packages/chains/src/common/docs.ts
```diff
@@ -0,0 +1,27 @@
+export const defaultDocs = (blockExplorerUrl: string, tokenAddress: string): string => {
+  return `Check out the token tracker for this asset in the 
+  <a href="${blockExplorerUrl}/token/${tokenAddress}" target="_blank" style="color: #BCAC83; cursor="pointer">Official Block Explorer</a>, where
+  you can access the token's site as well as market information`;
+};
+
+export const ellipsisDocs = (poolAddress: string, poolName: string, tokenAddress: string) => {
```
peetzweg, line 7: Maybe we can use some
```
<ol>
 <li> Step one </li>
</ol>
```

peetzweg: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol

NikMidas:
Tooltip Overlay Style: Slightly thicker border and the tooltip overlay should be slightly transparent

Formatting Style:
<ins>Deposit and Stake Instructions:</ins>
1) Receive LP Token at this [pool](url)
2) Deposit LP Token in our Midas [pool](url)

2022-08-17 at 7 35 AM

PR #515, opened 2022-08-17
Save Native Price in 5 Minute Intervals

### LINKED ISSUE #507: Store USD Price of Native Assets for further Distribution
## Description

We currently get the USD Price of the native assets of the network we are connected to, directly in the frontend from coingecko. Thereby making a bunch of requests from each client to coingecko, which sometimes leads to the user beeing blocked by coingecko.

https://github.com/Midas-Protocol/monorepo/blob/62761db5043f2154a75f971f4425fb4a3f6dd826/packages/ui/hooks/useUSDPrice.ts#L4-L27

Since the UI becoming cross chain, these requests will just become even more. Therefore we should reguarly pull the USD price of the native assets and deliver these via our custom backend from nextjs.

This might also come in handy in the new cross chain ui backend api calls.

## How

Although not ideal, we should make use of the already setup supabase connection with a new table to store the price data. I would suggest we don't store any historical data and just `upsert` data as needed.

peetzweg:
For convenient consumption for the backend and future frontend application.

TVL and Pool data, directly with USD pricing. Saving hassle in frontend to resolve multiple prices of native assets via coingecko.

Resolves #507.


packages/ui/hooks/useUSDPrices.ts
```diff
@@ -0,0 +1,27 @@
+import axios from 'axios';
+import { useQuery } from 'react-query';
+
+async function getUSDPriceOf(cgIds: string[]): Promise<number[]> {
```
carlomazzaferro, line 4: Shouldn't we fetch it from supabase?

peetzweg: Yea maybe, not sure yet. It's a WIP i wanted to include. 
It depends on the data we get from the call. Trying to split these up into reasonable chunks.

PR #516, opened 2022-08-17
Direct Imports of Local Packages

### LINKED ISSUE #506: Ability to import more granularly from the `chains` package
For ui we need to make sure can import individual packages. Otherwise all the pages will just blow up with a 2.2MB chains packages.

We need to create multiple entry points to export more granular what we need.
https://nodejs.org/api/packages.html#package-entry-points

I tried that but wasn't able to finish it yet properly. It work in development but once building the ui it broke.

peetzweg:
Enables direct imports of files from workspace packages via tsconfig.

Resolves #506 for me.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,7 +15,10 @@
       "@ui/networkData/*": ["ui/networkData/*"],
       "@ui/styles/*": ["ui/styles/*"],
       "@ui/images/*": ["ui/public/images/*"],
-      "@ui/icons/*": ["ui/public/icons/*"]
+      "@ui/icons/*": ["ui/public/icons/*"],
+      "@chains/polygon/*": ["chains/src/polygon/*"],
+      "@chains/bsc/*": ["chains/src/bsc/*"],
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -15,7 +15,10 @@
       "@ui/networkData/*": ["ui/networkData/*"],
       "@ui/styles/*": ["ui/styles/*"],
       "@ui/images/*": ["ui/public/images/*"],
-      "@ui/icons/*": ["ui/public/icons/*"]
+      "@ui/icons/*": ["ui/public/icons/*"],
+      "@chains/polygon/*": ["chains/src/polygon/*"],
+      "@chains/bsc/*": ["chains/src/bsc/*"],
+      "@chains/moonbeam/*": ["chains/src/moonbeam/*"]
```
carlomazzaferro, line 21: should we also add the testnet chains?

carlomazzaferro (approved):
LG, just the comment

tsconfig.json
peetzweg: i've added chapel. I guess we might never need these though. But let's see, can't hurt.

PR #526, opened 2022-08-18
get usdprice from supabase

smartdevcom:
We are now getting `usdPrice` from supabase.


peetzweg:
Don't merge just yet. The prices in supabase are not updated correctly. So this will result in usage of outdated prices.

peetzweg:
We are not doing this anytime soon.

Issue #357, opened 2022-07-12
FEATURE:  Add T&C for users opening the Midas app


NikMidas:
<ins>**Please describe the reasoning for this feature?**</ins> _Please mark one of the the checkboxes below, then put the description below that, screenshots preferred._ <br>
[ ] _Is this to resolve a process problem, bug. defect, error or glitch?_ <br>
[X] _Is this for a new feature or improvement?_ <br>
***
Link to Midas T&C: https://docs.google.com/document/d/1qs9hY009HNXe3IzxJmw03OyJt4HBdiSedpqIhHuFz54/edit?usp=sharing

Need to add Terms and Conditions similar to what Fei Protocol uses when launching the app. As soon as you click on Launch App on https://fei.money, the T&C will pop up on the App landing page:

![2022-07-12 at 3 34 PM](https://user-images.githubusercontent.com/103433798/178590235-06c266fa-d8ee-4831-ac53-10341d3604b0.png)

<ins>**Describe the solution or new feature you'd like.**</ins>  _How can we improve our product?  What will bring more users to our protocol?_ <br>

Would like to add our own T&C's for Midas.

<ins>**Describe alternatives that would satisfy requirements.**</ins> _What's an alternative that could satisfy the same requirements I was looking for?  Will it be harder?  Will it require more resources?._ <br>

<ins>**Additional context.**</ins> <br> _Any information that's relevant that doesn't fit in the categories above?_ <br>

PR #550, opened 2022-08-23
terms and conditions modal added

adub671:
## Description

Created terms & conditions modal upon midas app initialization

## Type of change

- [ ] Refactoring
- [x] New feature (non-breaking change which adds functionality)

Terms and condition modal pops up upon entering Midas app. If you accept, it will create a cookie indicating you have accepted this. The cookie expiration is based on session, once the session is closed the user will have to reaccept the terms.

## Related Issue(s)

Fixes: https://github.com/Midas-Protocol/monorepo/issues/357

### DIFF
--- a/packages/chains/deployments/ganache.json
+++ b/packages/chains/deployments/ganache.json
@@ -3,7 +3,7 @@
   "chainId": "1337",
   "contracts": {
     "AddressesProvider": {
-      "address": "0x5B3639BaDD3A08da48cBfb8F8451ff0035d9a4c8",
+      "address": "0xBd2c938B9F6Bfc1A66368D08CB44dC3EB2aE27bE",
       "abi": [
         {
           "anonymous": false,
@@ -380,7 +380,7 @@
       ]
     },
     "AddressesProvider_Implementation": {
-      "address": "0xb9e1c2B011f252B9931BBA7fcee418b95b6Bdc31",
+      "address": "0x2a504B5e7eC284ACa5b6f49716611237239F0b97",
       "abi": [
         {
           "anonymous": false,
@@ -613,7 +613,7 @@
       ]
     },
     "AddressesProvider_Proxy": {
-      "address": "0x5B3639BaDD3A08da48cBfb8F8451ff0035d9a4c8",
+      "address": "0xBd2c938B9F6Bfc1A66368D08CB44dC3EB2aE27bE",
       "abi": [
         {
           "inputs": [
@@ -762,7 +762,7 @@
       ]
     },
     "CErc20Delegate": {
-      "address": "0x6475789448864C5FA6fA2714331E4352b4187289",
+      "address": "0x9FBDa871d559710256a2502A2517b794B482Db40",
       "abi": [
         {
           "inputs": [],
@@ -2219,7 +2219,7 @@
       ]
     },
     "CErc20PluginDelegate": {
-      "address": "0x0689f3F569fdEff58967d0F4e41A690aa759596e",
+      "address": "0x2C2B9C9a4a25e24B174f26114e8926a9f2128FE4",
       "abi": [
         {
           "anonymous": false,
@@ -3729,7 +3729,7 @@
       ]
     },
     "CErc20PluginRewardsDelegate": {
-      "address": "0xE622C3c867FC0Dbc264E4b9D4c92586f1C8D80cf",
+      "address": "0x30753E4A8aad7F8597332E813735Def5dD395028",
       "abi": [
         {
           "anonymous": false,
@@ -5264,7 +5264,7 @@
       ]
     },
     "CEtherDelegate": {
-      "address": "0x0942Fb84dF77A1Ae96Af3903A5a245995e7fAEf1",
+      "address": "0xFB88dE099e13c3ED21F80a7a1E49f8CAEcF10df6",
       "abi": [
         {
           "inputs": [],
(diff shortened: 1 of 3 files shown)


peetzweg:
Great PR so far!
Here just some feedback for the next 👍 

+ make sure the code you commit is formatted as expected by 'prettier' (https://prettier.io/). If you are using VSCode, installing the prettier extensions should do the trick to fix what is off: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode
=> this enforces a unified style of source code throughout the whole codebase.

+ make sure you don't have any linter warnings/errors. We are using eslint (https://eslint.org/). It's similar to prettier, however, it does checks the code for bad patterns and marks them as warnings and errors. A extension should help you here as well:
https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

+ make sure all automated actions are running successful. If not make sure to fix them. Before that we will not merge anything. It can error on formatting problems (prettier) or lint problems (eslint) but also any other errors.
Screenshot 2022-08-24 at 11 55 21

+ you can probably revert the `ganache.json` file. It contains changes to the smart contracts deployed on your local machine. Feel free to ignore it for now. For most of us it will change all the time as we are not using the same 'private key' to deploy on the local blockchain node.

packages/chains/deployments/ganache.json
```diff
@@ -3,7 +3,7 @@
   "chainId": "1337",
   "contracts": {
     "AddressesProvider": {
-      "address": "0x5B3639BaDD3A08da48cBfb8F8451ff0035d9a4c8",
```
peetzweg, line 6: This whole file can be reverted. It will always change for you as you are using a different private key.

packages/ui/components/pages/Fuse/Modals/Terms.tsx
```diff
@@ -0,0 +1,168 @@
...
+    ModalContent,
+    ModalHeader,
+    ModalFooter,
+    ModalBody,
+    ModalCloseButton,
+    Text,
+  } from '@chakra-ui/react'
+
+  import React from 'react';
+  import { useState, useEffect } from 'react';
```
peetzweg, line 17: Fix prettier and eslint errors and warnings.

See suggested plugins in comment on pr.

`yarn workspace @midas-capital/ui lint`

packages/ui/components/pages/Fuse/Modals/Terms.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const h1 = {as: 'u', fontSize: 'xl', fontWeight: 'bold'};
+    const h2 = {as: 'u', fontSize: 'l', fontWeight: 'normal'};
+
+    return (
+        <Modal isOpen={isOpen} closeOnOverlayClick={false} onClose={accept} scrollBehavior='inside' size={'6xl'}  >
+          <ModalOverlay />
+          <ModalContent>
+            <ModalHeader>Terms & Conditions</ModalHeader>
+            <ModalBody>
+              <Text as={h1.as} fontSize={h1.fontSize} fontWeight={h1.fontWeight}>  1. Acceptance of Terms </Text>
```
peetzweg, line 40: If you like to use `<h1>` keep in mind that every single html page should only have a single `<h1>` element for accessibility reasons, screen readers, search engines etc.

So we should make use of all heading elements `<h1>,<h2>,<h3>,<h4>...`

best practice at least.

peetzweg: Just saw you defined `h1` yourself as an object on top. 
Was confused by the name. It works sure.

Chakra ui, the component framework we are using offers `Heading` and `Text` components. Better make sure to use those here.

Why should the elemnt be a `u` instead of a `p` which is probably used by `Text`?

packages/ui/components/pages/Fuse/Modals/Terms.tsx
```diff
@@ -0,0 +1,168 @@
...
+  import { useState, useEffect } from 'react';
+
+
+  const Terms = () => {
+
+    const [hasAcceptedTerms, setHasAcceptedTerms] = useState<boolean>(Boolean(document.cookie.split('; ').find((row) => row.startsWith('MidasTandCAccepted='))?.split('=')[1]));
+
+    let isOpen = !hasAcceptedTerms;
+    const accept = () => {
+      document.cookie = 'MidasTandCAccepted=true;';
```
peetzweg, line 26: Using a `cookie` works here but I think `localStorage` would be the way to go here. Cookies get sent back and forth between front & backend. This is not any of the requirement we need here.

https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

packages/ui/components/pages/Fuse/Modals/Terms.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+                  <ListItem>Intellectual Property Infringement: Engage in transactions involving items that infringe or violate any copyright, trademark, right of publicity or privacy or any other proprietary right under the law, including but not limited to sales, distribution, or access to counterfeit music, movies, software, or other licensed materials without the appropriate authorization from the rights holder; use of Midas Labs Ltd. & Midas Labs FZE LLC intellectual property, name, or logo, including use of Midas Labs Ltd. & Midas Labs FZE LLC trade or service marks, without express consent from Midas Labs Ltd. & Midas Labs FZE LLC or in a manner that otherwise harms Midas Labs Ltd. & Midas Labs FZE LLC; any action that implies an untrue endorsement by or affiliation with Midas Labs Ltd. & Midas Labs FZE LLC.</ListItem>
+                </UnorderedList>
+              </Text>
+            </ModalBody>
+            <ModalFooter>
+              <Button bg={'green'} onClick={accept}>Accept</Button>
```
peetzweg, line 160: use `colorScheme` here instead of changing the brackground color to predefined red and green html values. Looks pretty ugly :D 

https://chakra-ui.com/docs/components/button#button-colors

PR #565, opened 2022-08-26
Change the deployer

vminkov:
## Description

- tasks to change the deployer with a new one
- contracts deployments cleanup
- tweaked the config a bit so the UI can be used with mainnet forking


carlomazzaferro (approved):
Lovely stuff

carlomazzaferro:
I think the only reason this is failing is the `midasSdk.ts`'s `getOrCreateMidas()` needs to get the `InitializableClones` removed from it

PR #566, opened 2022-08-26
unsupported assets

smartdevcom:
unsupported assets by its property 'disabled`


peetzweg:
Not sure why we added this. We should not filter with a list in the frontend. Either we unsupport it in the comptroller or not.
This has been added here:
https://github.com/Midas-Protocol/monorepo/pull/368

Removing assets via a list just in the frontend causes calculation issue in the frotnend.

smartdevcom:
You mean `HIDE_ASSET` functionality?

PR #574, opened 2022-08-30
Show Cross Chain TVL

peetzweg:
This PR includes some rework regarding the SDK. This was needed in order to instantiate the SDK  and properly use the SDK without a proper Signer and only a Provider.
Provider => only connected to an RPC url, resolving contract interaction (No user address)
Signer => same as Provider but can also Sign actual transactions to be sent and included in the blockchain. (Includes user address)

So I wanted to be able to create the MidasSDK just with Providers and fail on functions which needed a signer. We have a lot of functions which don't care about a Signer. All of our read information functions. 
However, as a Provider has no "Address" attached to it, we need to make sure we use the `CallOverrides` prop to set a "Sender Address" to read functions which return a value based on the msg.sender .  Furthermore, we can only to `callStatic` contract interactions with a Provider only.

I've created a new "MultiMidas" context which stores an SDK instance for each supported chain. These sdks can be accessed to do crosschain queries from react-queries for example. See `useTVL` for details. 
It works, I'm not yet too happy with it. I guess over the time we use it we will refine it more. 

The useTVL mentioned is also the only usecase which uses this new crosschain data fetching. As soon as this is approved and merged we can tackle the bit more complex case of fetching the pools crosschain.

demo


packages/chains/src/chapel/params.ts
```diff
@@ -5,7 +5,7 @@ import chainAddresses from "./addresses";
 
 const specificParams: ChainParams = {
   blocksPerYear: BigNumber.from((20 * 24 * 365 * 60).toString()),
-  cgId: "binancecoin",
+  cgId: "binancecoin-chapel",
```
peetzweg, line 8: Had to change this, otherwise it clashes with the cgId of the actual BNB chain. It needs to be unqiue so it's not overriding BNB here:

https://github.com/Midas-Protocol/monorepo/blob/318683f1b3297ff7d87e99b1e3457ba616ff7b90/packages/ui/hooks/useUSDPrices.ts#L5-L11

packages/sdk/src/modules/FusePools.ts
```diff
@@ -18,16 +18,16 @@ export type LensPoolsWithData = [
 
 export function withFusePools<TBase extends MidasBaseConstructor>(Base: TBase) {
   return class FusePools extends Base {
-    async fetchFusePoolData(poolId: string, address?: string): Promise<FusePoolData> {
+    async fetchFusePoolData(poolId: string, overrides?: CallOverrides): Promise<FusePoolData> {
```
carlomazzaferro, line 21: we'll have to be careful here with our dependent libs. The bots for instance call this function a lot, so might have to check if their usage is affected

packages/sdk/tasks/one-time/arrakis-polygon-plugins.ts
```diff
@@ -128,7 +128,7 @@ task("arrakis:polygon:set-flywheels", "set plugin for each market")
       }
       try {
         //   Step 6: add Flywheels to market
-        await sdk.addFlywheelCoreToComptroller(mimoFlywheelAddress, COMPTROLLER, { from: signer.address });
+        await sdk.addFlywheelCoreToComptroller(mimoFlywheelAddress, COMPTROLLER);
```
carlomazzaferro, line 131: In line with these, this task must be updated as well  https://github.com/Midas-Protocol/monorepo/blob/development/packages/sdk/tasks/market/deploy-dynamic-rewards-market.ts

packages/ui/context/MultiMidasContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+  sdks: Partial<Record<SupportedChains, MidasSdk>>;
+}
+
+export const MultiMidasContext = createContext<MultiMidasContextData | undefined>(undefined);
+
+interface MultiMidasProviderProps {
+  children: ReactNode;
+}
+
+const ENABLED_CHAINS: SupportedChains[] = [56, 137];
```
carlomazzaferro, line 18: This is pretty cool. Should we add Moonbeam already, and also, use the @types module to get the ChainIds?

packages/chains/src/chapel/params.ts
carlomazzaferro: lovely solution :)

packages/sdk/src/modules/FusePools.ts
carlomazzaferro: Actually, looks like you've already made the necessary changes in the liquidations module, so we're good to go!

carlomazzaferro (approved):
Pending test fixes and discussions

smartdevcom (approved):

PR #582, opened 2022-08-31
added apr for aBNBc in Ankr

smartdevcom:
image

Added additional apr for `aBNBc` in Ankr. Contract `address` and `abi` is hardcoded in constant directory


packages/ui/constants/index.ts
```diff
@@ -58,3 +58,468 @@ export const MINUTES_PER_YEAR = 24 * 365 * 60;
 
 export const UP_LIMIT = 0.005;
 export const DOWN_LIMIT = 0;
+
+// for additional APR for aBNBC in Ankr
+export const aBNBcContractAddress = '0xBb1Aa6e59E5163D8722a122cd66EBA614b59df0d';
+export const aprDays = 7;
+export const aBNBcContractABI = [
```
carlomazzaferro, line 65: I think we can keep in the ABI here just the method we need, instead of having the whole thing

packages/ui/constants/index.ts
```diff
@@ -58,3 +58,468 @@ export const MINUTES_PER_YEAR = 24 * 365 * 60;
 
 export const UP_LIMIT = 0.005;
 export const DOWN_LIMIT = 0;
+
+// for additional APR for aBNBC in Ankr
+export const aBNBcContractAddress = '0xBb1Aa6e59E5163D8722a122cd66EBA614b59df0d';
```
carlomazzaferro, line 63: I believe this address is also availabe in the sdk, in the `assets` for tokenSymbol: `aBNBc`

smartdevcom: This is not a token address for `aBNBc`

carlomazzaferro: you're right, ignore

peetzweg:
👍 However, let's not make the apy bold. And treat it like any other apy we show. The only bold APY is the Supply APY.

packages/ui/constants/index.ts
smartdevcom: Done

smartdevcom:
> +1 However, let's not make the apy bold. And treat it like any other apy we show. The only bold APY is the Supply APY.

Done

PR #598, opened 2022-09-02
Consider Underlyings Decimals before Applying Native Prices

peetzweg:
convert to numbers considering the assets decimals before multiplying the price.

Screenshot 2022-09-02 at 12 23 15

### DIFF
--- a/packages/sdk/src/modules/FusePools.ts
+++ b/packages/sdk/src/modules/FusePools.ts
@@ -87,33 +87,38 @@ export function withFusePools<TBase extends MidasBaseConstructor>(Base: TBase) {
         );
 
         asset.supplyBalanceNative =
-          Number(utils.formatUnits(asset.supplyBalance)) * Number(utils.formatUnits(asset.underlyingPrice));
+          Number(utils.formatUnits(asset.supplyBalance, asset.underlyingDecimals)) *
+          Number(utils.formatUnits(asset.underlyingPrice));
 
         asset.borrowBalanceNative =
-          Number(utils.formatUnits(asset.borrowBalance)) * Number(utils.formatUnits(asset.underlyingPrice));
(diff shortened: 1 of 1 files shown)


packages/sdk/src/modules/FusePools.ts
```diff
@@ -87,33 +87,38 @@ export function withFusePools<TBase extends MidasBaseConstructor>(Base: TBase) {
...
-        const assetLiquidity =
-          Number(utils.formatUnits(asset.liquidity)) * Number(utils.formatUnits(asset.underlyingPrice));
 
         totalSuppliedNative += asset.totalSupplyNative;
         totalBorrowedNative += asset.totalBorrowNative;
 
-        asset.liquidityNative = assetLiquidity;
+        const assetLiquidityNative =
+          Number(utils.formatUnits(asset.liquidity, asset.underlyingDecimals)) *
+          Number(utils.formatUnits(asset.underlyingPrice));
```
carlomazzaferro, line 118: Is it missing here the `asset.underlyingDecimals`?

carlomazzaferro:
> convert to numbers considering the assets decimals before multiplying the price. Prices are still off, I guess because the oracles are not providing a correct price yet?
> 
> Screenshot 2022-09-02 at 12 23 15

These prices make some sense, but they are definitely not correct. Maybe I did some wrong math in the deploy script? https://github.com/Midas-Protocol/monorepo/blob/development/packages/sdk/chainDeploy/testnets/neondevnet.ts#L61

Doesn't seem like it honestly

peetzweg:
This also breaks the mimo pool on polygon.
But i'm quite sure this is the way to go. It seems to work correctly on all BSC pools though. 



Screenshot 2022-09-02 at 16 56 00



As far as I understand all values returned by `FusePoolLens.getPoolAssetsWithData()` are based on the `underlying.decimals()` except the price.

PR #628, opened 2022-09-07
dappeteer test for createPoolPage

smartdevcom:


peetzweg:
Does this work? Haven't checked it in more detail yet. How was the experience, should we use it? :)

smartdevcom:
Yes, it's much better than cypress-metamask-v2 or Synpress. closed to done of testing `CreatePoolPage`

packages/ui/.env.dist
```diff
@@ -16,3 +16,5 @@ SUPABASE_URL=https://xdjnvsfkwtkwfuayzmtm.supabase.co
 SUPABASE_PLUGIN_TABLE_NAME=apy-plugin
 SUPABASE_FLYWHEEL_TABLE_NAME=apy-flywheel
 SUPABASE_NATIVE_PRICES_TABLE_NAME=native-prices
+TEST_SEED=
```
peetzweg, line 19: To what values do I need to set these?

packages/ui/.env.dist
```diff
@@ -16,3 +16,5 @@ SUPABASE_URL=https://xdjnvsfkwtkwfuayzmtm.supabase.co
...
+TEST_PASS=
```
peetzweg, line 20: To what values do I need to set these?

peetzweg:
So I just run?
`yarn install`

`yarn test:ui` 

Can we run them in CI as well?

smartdevcom:
I think we should add it in CI check

peetzweg:
This is great stuff. I adapted the create pool test a bit to work with BSC local fork, to not create more and more testnet pools on chapel.

Creation is a great thing to test, we should also test other admin function.

However, to get a feel we could just start with checking that data is actually loading in, like balances and pool information.

Can we rewrite the MIMO test to work for example with the BSC jarvis version? SO all tests run on BSC Fork for now?

@carlomazzaferro mentioned he has issues with the Polygon fork becoming out of sync fast.

smartdevcom:
Okay, I will rewrite MIMO pool test in BSC forked

smartdevcom:
After I deposited some bnb on BSC forked, supplied value doesn't get updated so test couldn't be passed. And also creating a pool is not working for me, Did they work on your side? cc @peetzweg

docker-compose.yaml
```diff
@@ -7,7 +7,8 @@ services:
       context: .
     environment:
       - MNEMONIC
-      - FORK_URL_BSC
+      - FORK_URL_BSC # Leaving this in here, to potentially not break things?
```
peetzweg, line 10: @carlomazzaferro can you check this. I renamed this variable, maybe we need to adapt github secrets or other deployment environment variables somewhere?

packages/sdk/src/MidasSdk/index.ts
```diff
@@ -98,11 +100,45 @@ export class MidasBase {
     return this._signer;
   }
 
+  public get contracts(): {
```
carlomazzaferro, line 103: This will likely require changing all the bots code, AFAICT

carlomazzaferro: What's the reasoning here btw?

carlomazzaferro:
This is really dope. Just note my comment -- it'll require changing a bit the tests and the code for accessing the contracts, as that was refactored. Otherwise, looks really really awesome.

carlomazzaferro:
@peetzweg : this is the only place where the refactoring is needed: https://github.com/Midas-Protocol/monorepo/blob/development/.github/workflows/package-sdk-pull-request-build-and-test.yml#L67

carlomazzaferro (approved):
Pending the Mocha tests and the refactorings needed to fix those, LGTM

packages/sdk/src/MidasSdk/index.ts
peetzweg: probably not, it's a getter function. `sdk.contract.FusePoolLens` works as before, no worries.

I had to do it as somehow the instance created in the constructor gets out of sync and is not updated. I don't like this implementation as well, but it solves this issue. Wasn't able to find a better solution in a reasonable amount of time.

PR #640, opened 2022-09-08
added vault details in plugin tooltip & pointed strategy address

smartdevcom:
`Vault details` added
image

I am not sure the methods to point strategy address
Option1) Showing address beside of strategy name in the tooltip like below
image

Option2) Link to BSC scan with strategy address when clicking strategy name in the tooltip
image

Which way should I go? cc @carlomazzaferro  @peetzweg

### DIFF
--- a/packages/ui/components/pages/Fuse/FusePoolPage/SupplyList/AssetSupplyRow.tsx
+++ b/packages/ui/components/pages/Fuse/FusePoolPage/SupplyList/AssetSupplyRow.tsx
@@ -227,35 +227,34 @@ export const AssetSupplyRow = ({
                   placement="top-start"
                   body={
                     <Text lineHeight="base">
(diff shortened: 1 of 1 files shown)


peetzweg:
You can leave out the strategy address, it's not relevant for users. It would be nice for debugging but annoying anyone else.

The `[]` can be removed, it was just to indicate that it was using a link in markdown style.

`Read more about it in our Docs` can also be removed. The simpel link to the vault details if available is sufficient. 
If no vault link is available, link the docs as before: `Read more about it in our Docs`

smartdevcom:
Check again cc @peetzweg 

image

image

peetzweg:
Thats a great addition to show the address in case the plugin is not yet named 👍

peetzweg (approved):

PR #659, opened 2022-09-12
Fix `Sending a transaction requires a signer` #650

### LINKED ISSUE #650: BUG: "Sending a transaction requires a signer" for certain functions
Image

SDK refactoring issue: "Sending a transaction requires a signer". This has happened a few times in the SDK codebase after the initial multi-chain refactor -- we should double check how we are instantiating the SDK and the contracts, especially whenever a signer is needed.

peetzweg:
Previous to multichain sdk we had always been using the signer. This PR fallsback to signer by default.
I used provider in the beginning as I think it's best practice to assume read access by default and require write access just on demand. But this is a greater refactor.

So instantiate contracts with signer instead of provider only by default.

Fixes #650

### DIFF
--- a/packages/sdk/src/modules/CreateContracts.ts
+++ b/packages/sdk/src/modules/CreateContracts.ts
@@ -16,7 +16,7 @@ import { SignerOrProvider, SupportedProvider } from "../MidasSdk";
 export function withCreateContracts<TBase extends MidasBaseConstructor>(Base: TBase) {
   return class CreateContracts extends Base {
     createContractInstance<T extends Contract>(contract: keyof Artifacts) {
-      return (address: string, signerOrProvider: SignerOrProvider = this.provider) =>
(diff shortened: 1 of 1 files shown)


carlomazzaferro:
How can we be sure that the SDK gets instantiated in the UI with a signer?

carlomazzaferro (approved):

peetzweg:
> How can we be sure that the SDK gets instantiated in the UI with a signer?

As of now, we have multiple SDKs instatiated in the UI. 
We have the sdk from `useMidas / MidasContext` which we always have been using, this is an SDK which is always created with a signer and it's taking the signer from WAGMI.

The other sdks are `useMultiMidas / MultiMidasContext` which are just created with providers, so no signer here.

We don't use the "MultiMidas" yet anywhere else except for TVL. I will probably add more refactors once I have more focus time on getting back working on fetching pools from multiple chains, PR.

PR #661, opened 2022-09-12
min borrow should be bigger

smartdevcom:
Make min borrow value a little bigger than real one

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "dev:node": "yarn workspace @midas-capital/sdk hardhat node",
     "dev:fork:chapel:node": "FORK_URL_BSC=${MY_CHAPEL_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=97 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
     "dev:fork:bsc:node": "FORK_URL_BSC=${MY_BSC_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=56 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
+    "dev:fork:polygon:node": "FORK_URL_POLYGON=${MY_POLYGON_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=137 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "dev:node": "yarn workspace @midas-capital/sdk hardhat node",
     "dev:fork:chapel:node": "FORK_URL_BSC=${MY_CHAPEL_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=97 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
     "dev:fork:bsc:node": "FORK_URL_BSC=${MY_BSC_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=56 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
+    "dev:fork:polygon:node": "FORK_URL_POLYGON=${MY_POLYGON_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=137 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
```
peetzweg, line 49:
```suggestion
    "dev:fork:polygon:node": "FORK_URL_BSC=${MY_POLYGON_FORK_URL} FORK_BLOCK_NUMBER= FORK_CHAIN_ID=137 yarn workspace @midas-capital/sdk hardhat node --no-deploy --tags fork",
```

The env variable used throughout hh config and docker is called `FORK_URL_BSC` so it's branded with BSC, which is stems from the first time we introduced fork. 
When we set `FORK_URL_POLYGON` it's not used anywhere. So we need to use  `FORK_URL_BSC` instead.

packages/ui/utils/formatNumber.ts
```diff
@@ -20,3 +20,7 @@ export const bigMul = (value1: string, value2: string) => {
   const val = multiply(bignumber(value1), bignumber(value2));
   return format(val, { notation: 'fixed' });
 };
+
+export const toCeil = (value: number, len: number) => {
```
peetzweg, line 24: Nice. 👍 
Sad thought that we don't have a test suite to add some unit test for this function.

package.json
smartdevcom: fixed

packages/ui/utils/formatNumber.ts
smartdevcom: I will add it later on

PR #679, opened 2022-09-13
added sourceMap options in tsconfig and nextconfig

adub671:
## Type of change

- [ ] Refactoring
- [ x] New feature (non-breaking change which adds functionality)

Added source map configs for Typescript compiler and Next config file. Was unable to verify locally; this may be something we can only see in production

Used these resources to find source map configs: 
[Next Source Map Documentation](https://nextjs.org/docs/advanced-features/source-maps)
[TypeScript Source Map Documentation](https://www.typescriptlang.org/tsconfig#sourceMap)

## Related Issue(s)

Fixes

### DIFF
--- a/packages/ui/tsconfig.json
+++ b/packages/ui/tsconfig.json
@@ -17,8 +17,8 @@
     "incremental": true,
     "downlevelIteration": true,
     "composite": true,
-    "sourceMap": true,
+    "sourceMap": true
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
-  "exclude": ["api", "node_modules"],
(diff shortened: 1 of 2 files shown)


packages/ui/tsconfig.json
```diff
@@ -16,8 +16,9 @@
     "jsx": "preserve",
     "incremental": true,
     "downlevelIteration": true,
-    "composite": true
+    "composite": true,
+    "sourceMap": true,
```
carlomazzaferro, line 20: trailing commas are invalid in json files, please do remove them here and below!

carlomazzaferro (changes requested):
see comment

Issue #653, opened 2022-09-12
Calculation of limits is off


carlomazzaferro:
Neon devnet -- all the prices look good, but the borrow limits are 3 decimal points off.

I can verify that the price oracle returns the correct value for each of the assets's prices. 

Neon devnet's tokens can be found here: https://neonscan.org/tokens

![Image](https://user-images.githubusercontent.com/9201443/189581832-7c20a28e-89c2-47ea-958a-6d24bafb69c4.png)

This also causes the issue of not being able to deposit as much as needed (560 * 0,05 = 28$ , which is less than 100$).

![Image](https://user-images.githubusercontent.com/9201443/189582189-e14d12dd-6677-4ba4-8a5d-f08aabbea341.png)

smartdevcom:
Check https://testnet.midascapital.xyz/245022926/pool/0 again, I could see it is working.

![image](https://user-images.githubusercontent.com/45715420/189601179-869393df-6e7a-4cf2-b651-dd09ee0bd658.png)

carlomazzaferro:
This is also concerning. @peetzweg are you able to repro? I still get 560 as the limit

peetzweg:
maxBorrows return value seems to be always scaled by 6, instead of the underlyings decimals.


Comptroller: `0x74222882649f69eEdaeCc86e356B40534B27BDF2`

wBTC ctoken: `0x0F1Bf4237861a129ecDf8a3360DD3d970D07a905` 8 decimals
WNEON cToken: `0x4962a43589B6eAC2Bf89F83c8d02C6d3e6B06e2E` 12 decimals
ETH ctoken: `0x0F1Bf4237861a129ecDf8a3360DD3d970D07a905` 6 decimals 
USDC ctoken: `0x37De21d2e5f29CA67f9D3F4360a5257Da05Ae63E` 6 decimals

NikMidas:
Per Peet, this can be tested and possibly reproduced in Foundry

peetzweg:
So it only occurs in a pool with `ctokens` which have different `decimals`. The only one in production I was able to find is: 

MIMO<>Midas https://app.midascapital.xyz/137/pool/3
and
Midas<>NEON https://testnet.midascapital.xyz/245022926/pool/0

carlomazzaferro:
closing in favor of https://github.com/Midas-Protocol/monorepo/issues/733

smartdevcom:
resolved again

PR #684, opened 2022-09-14
Fix/653

peetzweg:
Refactors a lot about assets passing. Not clean yet but see comment what went wrong.

Related to #653 but does not fix it finally. There seems to be an issue with the value we get from `FusePoolLensSecondary.getMaxBorrow` on Neon, or even on the other networks as well if the some condition is met. It's a bit weird. Need to investigate further tomorrow.

https://github.com/Midas-Protocol/contracts/blob/88b4e976c362a12059f5d8fd1a3a7a19d8d32e6a/contracts/FusePoolLensSecondary.sol#L136-L144


packages/ui/components/pages/Fuse/FusePoolPage/BorrowList/index.tsx
```diff
@@ -113,37 +113,37 @@ export const BorrowList = ({ assets, borrowBalanceFiat, comptrollerAddress }: Bo
         <Tbody>
           {assets.length > 0 ? (
             <>
-              {borrowedAssets.map((asset, index) => {
+              {borrowedAssets.map((asset) => {
                 return (
                   <AssetBorrowRow
                     comptrollerAddress={comptrollerAddress}
                     key={asset.underlyingToken}
-                    assets={borrowedAssets}
```
peetzweg, line 121: In these lines only a slice of the actual `assets` were passed. Which causes some calculations later down the line to be off, as they did not have access to all of the assets of the pool.

smartdevcom (approved):

PR #695, opened 2022-09-16
merged supply and borrow list

smartdevcom:
image

Now `Supply` and `Borrow` lists are merged and `Asset Details` is moved to additional panel of each asset row.
Merged assets list can be filtered, sorted.


packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetModal.tsx
```diff
@@ -135,7 +135,7 @@ const AddAsset = ({ comptrollerAddress, onSuccess, poolID, poolName }: AddAssetP
               poolData.assets.map((asset, index) => {
                 return (
                   <WrapItem key={index}>
-                    <Button variant="solid" px={2}>
+                    <Button variant="_solid" px={2}>
```
peetzweg, line 138: Can't we use `solid` or adapt it? 
If we need to create a new style I personally would prefer a name without `_` underscore as it's indicating some variables to not mess/placeholder with instead of a `style` I should use all over the place

packages/ui/components/pages/Fuse/FusePoolEditPage/FlywheelEdit/index.tsx
```diff
@@ -101,7 +101,7 @@ const FlywheelEdit = ({ pool }: { pool: PoolData }) => {
             <Heading size="md">Flywheels</Heading>
 
             <Flex mt={{ base: 2, md: 0 }} ml="auto" flexWrap="wrap" gap={2}>
-              <Button variant="ghost" onClick={openAdd} ml="auto">
+              <Button variant="_ghost" onClick={openAdd} ml="auto">
```
peetzweg, line 104: Same as with `_solid`

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/AdditionalInfo.tsx
```diff
@@ -0,0 +1,239 @@
...
+        templateColumns={{
+          base: 'repeat(1, 1fr)',
+          lg: 'repeat(2, 1fr)',
+        }}
+        w="100%"
+        gap={4}
+        alignItems="flex-end"
+      >
+        <Box
+          height="200px"
```
peetzweg, line 90: Seeing hardcoded values here, how does it look on mobile? Seem like some of the users and from the team really look at the UI in mobile often.

packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetModal.tsx
smartdevcom: Well, chakra provides some default variants buttons and I need them to tag, if we use 'solid` as before, then its style would be overrided

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/BorrowApy.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { useMidas } from '@ui/context/MidasContext';
+import { useColors } from '@ui/hooks/useColors';
+import { MarketData } from '@ui/types/TokensDataMap';
+import { getBlockTimePerMinuteByChainId } from '@ui/utils/networkData';
+
+export const BorrowApy = ({ asset }: { asset: MarketData }) => {
+  const { currentChain, midasSdk } = useMidas();
+  const blocksPerMin = getBlockTimePerMinuteByChainId(currentChain.id);
+  const borrowAPR = midasSdk.ratePerBlockToAPY(asset.borrowRatePerBlock, blocksPerMin);
```
peetzweg, line 12: `borrowAPR` and `blocksPerMin` could be wrapped in a `useMemo()`.

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/FundButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+}: {
+  comptrollerAddress: string;
+  assets: MarketData[];
+  asset: MarketData;
+  mode: FundOperationMode;
+  isDisabled?: boolean;
+}) => {
+  const { isOpen: isModalOpen, onOpen: openModal, onClose: closeModal } = useDisclosure();
+
+  const modeName = useMemo(() => FUNDOPERATION_MODE_NAMES[mode], [mode]);
```
peetzweg, line 24: We could potentially get the name of the FundOperation from the Enum itself. We can assign strings to Enums which we could than retrieve, so no need for externally keeping the `FUNDOPERATION_MODE_NAMES`.

https://bobbyhadz.com/blog/typescript-convert-enum-to-string

packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetModal.tsx
peetzweg: Ah I see you changed all the variant names. But why again? It's a bit unclear to me.
Shouldn't using `solid` as before work?

Screenshot 2022-09-21 at 11 35 25

peetzweg:
Screenshot 2022-09-21 at 11 45 52

peetzweg:
The selection top row is shifting but only if selecting and deselecting the borrowable button.

demo

peetzweg:
Let's use `↓` and `↑` for sorting. The triangle is quite heavy. Maybe there are chakra icons for these arrows if we need an icon instead of a character.


Screenshot 2022-09-21 at 11 53 45

Also If the markets are sorted by default by either category anyways, can't we just show this indicator for this category by default? So the user can directly see how the markets are sorted?

peetzweg:
Pool details border is doubled.

Screenshot 2022-09-21 at 11 59 14 1

peetzweg:
Make these two border the same thickness when details are shown.

Screenshot 2022-09-21 at 12 01 41

The inners of the asset row are jumping when opening the details. Can we fix this easily?

demo

peetzweg:
On Mobile it looks broken. It does not have to be usable on mobile but at least it should look like it's working. Some people are showing of our app on their mobile phone.


Screenshot 2022-09-21 at 12 05 30


As before with the old system, we should prevent some columns from showing.
But maybe there is also a better way, maybe react-table has some docs for mobile layout?

Screenshot 2022-09-21 at 12 05 37

peetzweg:
Screenshot 2022-09-21 at 12 11 14

Euler Color for reference, it also has blue collateral. The Collateral APY is good, so blue is more friendly. Borrow APY is "bad", because you have to pay it, so better to be orange.

Screenshot 2022-09-21 at 12 14 46


Old version had TVL in Market column. We need to save on space, for future oracle data and mobile anyways. So somehow it would nice to fit in there again.

Screenshot 2022-09-21 at 12 16 08

packages/ui/components/pages/Fuse/FusePoolEditPage/FlywheelEdit/index.tsx
smartdevcom: https://github.com/Midas-Protocol/monorepo/pull/695#discussion_r976281301

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/AdditionalInfo.tsx
```diff
@@ -0,0 +1,239 @@
...
+  supplyBalanceFiat,
+}: {
+  row: Row<Market>;
+  rows: Row<Market>[];
+  comptrollerAddress: string;
+  supplyBalanceFiat: number;
+}) => {
+  const asset: MarketData = row.original.market;
+  const assets: MarketData[] = rows.map((row) => row.original.market);
+  const assetUtilization =
```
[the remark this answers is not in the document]
smartdevcom: Fixed!

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/AdditionalInfo.tsx
smartdevcom: it's graph height and it works both desktop and mobile

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/BorrowApy.tsx
smartdevcom: Fixed

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/FundButton.tsx
smartdevcom: Fixed

smartdevcom:
> Screenshot 2022-09-21 at 11 45 52

Fixed

smartdevcom:
> deselecting

Fixed

smartdevcom:
> Let's use `↓` and `↑` for sorting. The triangle is quite heavy. Maybe there are chakra icons for these arrows if we need an icon instead of a character.
> 
> Screenshot 2022-09-21 at 11 53 45
> 
> Also If the markets are sorted by default by either category anyways, can't we just show this indicator for this category by default? So the user can directly see how the markets are sorted?

Market's list isn't sorted first time because it's listed by collateral first

packages/ui/components/pages/Fuse/FusePoolPage/MarketsList/AdditionalInfo.tsx
peetzweg: Ah alright makes sense to give it a fixed height 👍

peetzweg:
@smartdevcom  can we make the LTV value more like it was before? So smaller and with LTV suffix. I think it a bit confusing if there is just a percentage in brackets.

Screenshot 2022-09-21 at 16 05 23

peetzweg:
Not sure if this can be changed per column. But if so this would be nice.

Screenshot 2022-09-21 at 16 04 18

peetzweg:
Can we move the sorting arrow to the front of the title?
It's shaking the table header if enabled. It could be prevented. We have enough space in the table header.


Screenshot 2022-09-21 at 16 00 49

demo

peetzweg:
Overall the font got bigger on every text. How could this have happen?
Looks nice but also a bit odd. Not sure if I just need to get used to it or if it's not merging in with the rest of the text sizes throughout the app.

Screenshot 2022-09-21 at 16 03 35

peetzweg:
The Reward Badge is bigger than the others. Can we make all have the same height?

Screenshot 2022-09-21 at 16 24 05

peetzweg:
I know it's a hard one but can we somehow give supply and borrow APY headers more room?
It would be nice to have them in a single line. 
For `Borrow Balance` and `Supply Balance` it's fine for me. But it looks so cramped for `Supply APY` and `Borrow APY`.


Screenshot 2022-09-21 at 16 25 33

packages/ui/components/pages/Fuse/FusePoolEditPage/AssetConfiguration/AddAssetModal.tsx
smartdevcom: I am using default `solid` variant buttons as filters like `Collateral`, `Protected`, `Borrowable`. and if we use as before, then `solid` variants are overrided with our style so couldn't use it

peetzweg:
@smartdevcom 
After supplying the asset jumps to the top and the asset with the open details changes.
Can we close the details of an asset when opening the Amount Modal?



demo

peetzweg:
@smartdevcom can we get the whole row clickable to open the details? That would be handy. It's not as fluid to use the botton on the right. Also saves more space again.

Screenshot 2022-09-22 at 17 11 18

smartdevcom:
> @smartdevcom After supplying the asset jumps to the top and the asset with the open details changes. Can we close the details of an asset when opening the Amount Modal?
> 
> demo [ demo ](https://user-images.githubusercontent.com/839848/191783629-52dbff37-f45a-45a6-aaa4-d3e7f34b8b8c.gif) [ ](https://user-images.githubusercontent.com/839848/191783629-52dbff37-f45a-45a6-aaa4-d3e7f34b8b8c.gif)

Otherwise we could stop sorting as collateral, so table row wouldn't be changed after it's being used as collateral. user could see the badge, switch and sort if he wants to see if asset is collateral or not, wdyt?

smartdevcom:
> @smartdevcom Seems like we are checking the wrong variable to determine the "REWARDS" badge and filter.
> 
> Screenshot 2022-09-22 at 16 22 18

I guess rewards banner seems to be wrong because these assets of your sceenshot don't have any deposited so there shouldn't be rewards yet. the rewards banner `This pool is offering rewards.....` means if pool has rewards totally or particular person has rewards in this pool?

smartdevcom:
Now at first, list is sorted by market and sorting icon is enabled, supply action is tested and it works as before without any flickering.

PR #700, opened 2022-09-20
available to withdraw replaced wallet balance in withdraw mode

adub671:
## Description

On withdraw modal, the Wallet Balance is changed to "Available To Withdraw" and fetches the max withdrawable balance

## Type of change

- [x] Docs change / dependency upgrade
- [ ] Refactoring

 On withdraw modal, the Wallet Balance is changed to "Available To Withdraw" and fetches the max withdrawable balance

## Related Issue(s)

Fixes


packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
```diff
@@ -240,6 +242,15 @@ const AmountSelect = ({
     }
   };
 
+  const updateAvailableToWithdraw = async () => {
+    const max = await fetchMaxAmount(mode, midasSdk, address, asset);
+    setAvailableToWithdraw(utils.formatEther(max));
```
smartdevcom: I think this max amount should be scaled by `UnderlyingDecimals`

smartdevcom (approved):

packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
```diff
@@ -240,6 +242,15 @@ const AmountSelect = ({
     }
   };
 
+  const updateAvailableToWithdraw = async () => {
```
peetzweg, line 245: Should be wrapped in a `useCallback()` hook to prevent the function from beeing created on every render.

https://reactjs.org/docs/hooks-reference.html#usecallback

smartdevcom: Fixed

smartdevcom:
Seems Ok to merge!

packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
adub671: fixes look good to me! tested & working

PR #701, opened 2022-09-20
abbreviate long token names

adub671:
## Description

Created an abbreviator function and applied to symbols in both Asset Supply Row & Asset Borrow Row. Popup Tooltip appears upon hover of abbreviated symbols

## Type of change

- [x] Docs change / dependency upgrade
- [ ] Refactoring

Pop up tool tip with full symbol name appears for characters longer than 9 characters for better formatting 

## Related Issue(s)

Fixes


adub671:
image
image

first screenshot is from PR. second is from my UI. I am only seeing the plugin symbol, I am not sure why this discrepancy exists

packages/ui/utils/abbreviator.tsx
```diff
@@ -0,0 +1,15 @@
+import { Text } from '@chakra-ui/react';
+import { ReactNode } from 'react';
+
+import { PopoverTooltip } from '@ui/components/shared/PopoverTooltip';
+
+export function abbreviator(str: string | undefined, len: number): string | undefined | ReactNode {
```
peetzweg, line 6:
```suggestion
export function abbreviator(str?: string, len: number): string | undefined | ReactNode {
```

peetzweg: I think if it's either the value or undefined using the `?` optional operator is a bit cleaner in my opinion. But we could keep it as is. ✌️

packages/ui/utils/abbreviator.tsx
```diff
@@ -0,0 +1,15 @@
...
+  if (str === undefined) return undefined;
+  return str?.length <= len ? (
+    str
+  ) : (
+    <PopoverTooltip placement="top-start" body={str}>
```
peetzweg, line 11: Interesting to insert a popover around the abbreviated text to show the original string.
I would be a bit cautious if it really works as expected in every part of the app, as I don't know the inner workings of the Popover at all.

peetzweg:
For reference there is also a CSS property to handle overflowing text:

https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

peetzweg:
> first screenshot is from PR. second is from my UI. I am only seeing the plugin symbol, I am not sure why this discrepancy exists

@adub671 so there is still a bug left? Maybe it's due to using the Popover? Maybe try it without the Popover. Using the primitive html `<span title="long text">abbreviated text</span>` should work for sure.

packages/ui/components/pages/Fuse/FusePoolPage/SupplyList/AssetSupplyRow.tsx
```diff
@@ -359,7 +360,9 @@ export const AssetSupplyRow = ({
                   Number(utils.formatUnits(asset.supplyBalance, asset.underlyingDecimals)) <
                     UP_LIMIT &&
                   '+'}{' '}
-                {tokenData?.extraData?.shortName ?? tokenData?.symbol ?? asset.underlyingSymbol}
+                {abbreviator(tokenData?.extraData?.shortName, 9) ??
```
adub671, line 363: @peetzweg are we still using `tokenData.extraData.shortname` 
I noticed this only exists in `AssetSupplyRow.tsx` and is omitted in `AssetBorrowRow.tsx`

Let me know if I should remove it or add it to borrow row

adub671:
features outdated due to new interface for supply and borrow row

PR #706, opened 2022-09-20
deploy to chapel artifacts

vminkov:
## Description

deploying fixes for the flywheels, max redeem amount and reduced comptroller code size


carlomazzaferro:
I suppose the tests are failing because of the comproller changes not being live, is that the right assumption? But why would local tests fail>?

PR #722, opened 2022-09-22
input value should be 2^256-1 when max withdraw

smartdevcom:


peetzweg:
Ah thanks for picking it up, will review this asap. 👍

peetzweg:
Works great LFG!

PR #729, opened 2022-09-23
692 show approve button before supply button

adub671:
## Description
Approve Button now appears if we need to approve before supplying to the pool. Added .approve() in FundOperations.ts in SDK & refactored .supply() and .repay() to call this function

## Type of change

- [ ] Refactoring
- [x] New feature (non-breaking change which adds functionality)

If approval to spend needed, user will first see the button "Approve" then after accepted transaction in metamask, it will show "Supply" upon which the user will need to accept the transaction to supply (mint)

## Related Issue(s)

Fixes


packages/sdk/src/modules/FundOperations.ts
```diff
@@ -39,11 +51,9 @@ export function withFundOperations<TBase extends MidasBaseConstructor>(Base: TBa
       const currentSignerAddress = await this.signer.getAddress();
 
       const hasApprovedEnough = (await token.callStatic.allowance(currentSignerAddress, cTokenAddress)).gte(amount);
-      if (!hasApprovedEnough) {
-        const max = BigNumber.from(2).pow(BigNumber.from(256)).sub(constants.One);
-        const approveTx = await token.approve(cTokenAddress, max);
-        await approveTx.wait();
-      }
+      // if (!hasApprovedEnough) {
```
peetzweg, line 54: If code is commented out without an explanation it should be just removed. We can restore everything from the git history if ever needed again.

packages/sdk/src/modules/FundOperations.ts
```diff
@@ -28,6 +28,18 @@ export function withFundOperations<TBase extends MidasBaseConstructor>(Base: TBa
       return { gasWEI, gasPrice, estimatedGas };
     }
 
+    async approve(cTokenAddress: string, underlyingTokenAddress: string) {
+      const token = getContract(underlyingTokenAddress, this.artifacts.EIP20Interface.abi, this.signer);
+      const max = BigNumber.from(2).pow(BigNumber.from(256)).sub(constants.One);
+      const tx = await token.approve(cTokenAddress, max);
```
peetzweg, line 34:
```suggestion
return token.approve(cTokenAddress, max);
```

Lets just return the actual transaction. This way the user of this function can decide what to do with it and how to handle the error.

In 39 the code just sends back a generic error code obfuscating the actual error code of the failed transaction.

packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
```diff
@@ -287,6 +293,11 @@ const AmountSelect = ({
     }
   }, [mode, updateAvailableToWithdraw]);
 
+  const token = getContract(asset.underlyingToken, ERC20Abi, midasSdk.signer);
```
peetzweg, line 296: Can't see the whole context here. Could maybe better be in a useEffect() hook or useMemo.

Otherwise it will be executed on each render, which is not good at all.

peetzweg (changes requested):

peetzweg:
@adub671 this PR still seems to have some unresolved comments. ✌️

packages/sdk/src/modules/FundOperations.ts
carlomazzaferro: Returns tx, resolved

carlomazzaferro:
> @adub671 this PR still seems to have some unresolved comments. ✌️

Is this still the case?

packages/ui/components/pages/Fuse/Modals/PoolModal/AmountSelect.tsx
adub671: this line is now in a useEffect hook

adub671:
> @adub671 this PR still seems to have some unresolved comments. ✌️

everything seems to be taken care of!

peetzweg:
We changed the whole flow in #849 . So this is obsolete.

PR #740, opened 2022-09-26
added deprecated feature

smartdevcom:
image
image


peetzweg:
Can we just not show this "Deprecated" filter at all, if there is no deprecated market visible to this user at all?

smartdevcom:
> Can we just not show this "Deprecated" filter at all, if there is no deprecated market visible to this user at all?
> 
> Will only be confusing and it will best case never be needed anyways.

Done

peetzweg:
Epic you even disabled the supply button 👍 

Screenshot 2022-09-26 at 17 35 17

smartdevcom:
It's also disabled in modal

image

PR #751, opened 2022-09-28
cross chain pools list

smartdevcom:


packages/ui/components/pages/Fuse/FusePoolsPage/FuseDashNav/CrossFilterButtons.tsx
```diff
@@ -0,0 +1,68 @@
+import { ButtonGroup, Img } from '@chakra-ui/react';
+import { ChainConfig } from '@midas-capital/types';
+import { Dispatch, useMemo } from 'react';
+
+import { CButton } from '@ui/components/shared/Button';
+import { SimpleTooltip } from '@ui/components/shared/SimpleTooltip';
+import { useMultiMidas } from '@ui/context/MultiMidasContext';
+import { getChainConfig } from '@ui/utils/networkData';
+
+export const CrossFilterButtons = ({
```
peetzweg, line 10: Let's call it `ChainFilterButtons`

smartdevcom: Okay, maybe it needs to be refactored by `react-table`

packages/ui/constants/index.ts
```diff
@@ -92,3 +92,9 @@ export const COLLATERAL_FACTOR_TOOLTIP =
   'Collateral factor can range from 0-90%, and represents the proportionate increase in liquidity (borrow limit) that an account receives by depositing the asset.';
 export const ADMIN_FEE_TOOLTIP =
   "The fraction of interest generated on a given asset that is routed to the asset's admin address as a fee.";
+export const ENABLED_CHAINS: SupportedChains[] = [
```
peetzweg, line 95: This value should change for `production` & `development`.

Can we maybe somehow reuse the netlify environment variables?

```
EVMOS = "true"
BSC = "true"
MOONBEAM = "true"
POLYGON = "true"
ARBITRUM = "true"
```

peetzweg: https://github.com/Midas-Protocol/monorepo/blob/ef1f83ceeb81022f233ccac6be2e27ec27b55870/netlify.toml#L20-L24

smartdevcom: makes sense, I will read it from `.env`

packages/ui/types/ChainMetaData.ts
```diff
@@ -38,3 +38,7 @@ export const supportedChainIdToConfig: {
   [chapel.chainId]: { enabled: true, supported: config.isDevelopment || config.isTestnetEnabled },
   [ganache.chainId]: { enabled: config.isDevelopment, supported: config.isDevelopment },
 };
+
+export interface CrossFusePools {
```
peetzweg, line 42:
```suggestion
export interface FusePoolsPerChain {
```

I'm a fan of this `Cross` naming. It think it's a bit misleading. The functionalty is not really "across" chains it's just showing data from multiple chains at once.

In my opinion `FusePoolsPerChain` is a bit better describing what the type is.

peetzweg:
Just fyi, I know this was a huge task but next time it would be nice to have it into smaller PRs. This touches 100 files. I can't review all the code changes 🙈

So maybe splitting up the changes in multiple steps, like basic required changes & redesign

peetzweg:
I guess this is unfinished? Looks broken. I'm fine if it just looks like the "Pool Details" before the use of react-table.

demo

smartdevcom:
> I guess this is unfinished? Looks broken. I'm fine if it just looks like the "Pool Details" before the use of react-table. Step by step. v
> 
> demo [ demo ](https://user-images.githubusercontent.com/839848/193041340-3b5f927a-1066-4b1f-8cfe-1973318cb217.gif) [ ](https://user-images.githubusercontent.com/839848/193041340-3b5f927a-1066-4b1f-8cfe-1973318cb217.gif)

Yes, that's not finished yet, will do soon

peetzweg:
Screenshot 2022-09-29 at 15 17 38

peetzweg:
can we make these rows able to be
"open in a new Tab"?
Maybe using (a) or a chakra link?
Rows Per Page :
I guess we are using a
'onclick' handler, this does not support
opening in tabs
Screenshot 2022-09-29 at 15 21 16

peetzweg:
Nice touch with the tooltip for `-` 👍 

Screenshot 2022-09-29 at 15 23 45

peetzweg:
Can we give the Pool Name column a fixed length or flex fraction so it's not resizing if there are longer pool names visible or not?

We could abbreviate too long pool names with the `text-overflow:ellipsis` css property
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow?retiredLocale=de

demo

peetzweg:
Skeleton needs to be updated.

Screenshot 2022-09-29 at 15 32 44

peetzweg:
Change label and move the page Stuff to the bottom.

Screenshot 2022-09-29 at 15 33 47


Do the same on the Pool Detail Page:

Screenshot 2022-09-29 at 15 35 59


Fix Page Number Prompt and hide it if only no additional pages available.

Screenshot 2022-09-29 at 15 34 27

peetzweg:
There is enough space for them to be horizontal on mobile. Looks odd if they go into this 2x2 grid.


Screenshot 2022-09-29 at 15 38 23

peetzweg:
Redirects of the chain select are linking to an empty route.

demo

peetzweg:
The left align columns headers are not aligned with the column content. Can this be fixed?

Screenshot 2022-09-29 at 15 41 08

peetzweg:
If not connected, sort the pools by `Total Supply`

Screenshot 2022-09-29 at 15 42 59

peetzweg:
As with Borrow and Supply balance. Always show 2 digits for Supply and Borrow APYs.

Screenshot 2022-09-29 at 15 45 04

peetzweg:
The flywheel rewards don't seem to be loading anymore. Or if they do, way slower than before?

Screenshot 2022-09-29 at 15 46 31

peetzweg:
The create pool button now is quite lost.
Let's move it to the top hero element.


Screenshot 2022-09-29 at 15 49 09

peetzweg:
Sorting by chain does not make sense.

demo

peetzweg:
The line height of the heading for `Supply APY`, `Borrow APY` is different to `Supply Balance` and `Borrow Balance`.

All of the column headers should have the same line height.

Screenshot 2022-09-29 at 15 47 52

peetzweg:
APYs are not showing. These should however even be visible even if no wallet is connected.

Screenshot 2022-09-29 at 15 59 38

peetzweg:
Please also always show 2 digits for the asset value in tokens.

Screenshot 2022-09-29 at 16 00 40

peetzweg:
Update placeholder texts.


Screenshot 2022-09-29 at 16 09 15

Screenshot 2022-09-29 at 16 10 40

peetzweg:
If not logged in, all the actions of a market should be disabled.
We could also show the `Connect Wallet` button in the additional details.

Screenshot 2022-09-29 at 16 15 13

smartdevcom:
image

peetzweg:
These values are not on the same baseline.

Screenshot 2022-09-30 at 12 41 20

packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList/index.tsx
```diff
@@ -10,11 +20,41 @@ const FusePoolList = () => {
...
   const { isLoading, poolsPerChain, error } = useCrossFusePools([...enabledChains]);
   const [err, setErr] = useState<Err | undefined>(error as Err);
+  const mounted = useRef(false);
+  const [preSearchText, setPreSearchText] = useState<string | undefined>(undefined);
+  const [preFilter, setPreFilter] = useState<(string | SupportedChains)[] | undefined>(undefined);
+  const [preSorting, setPreSorting] = useState<SortingState | undefined>(undefined);
+  const { address } = useMultiMidas();
 
   useEffect(() => {
     setErr(error as Err);
```
peetzweg, line 30: Why set `err` when `error` is changing? `error` could be used directly from the `useQuery` `useCrossFusePools` hook.

smartdevcom: This `err` is for each chain, so for example if `polygon` got error, then `bsc` should be shown, so we should show each chain individually

packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList/index.tsx
```diff
@@ -26,7 +66,15 @@ const FusePoolList = () => {
     );
   }
 
-  return <PoolsRowList poolsPerChain={poolsPerChain} isLoading={isLoading} />;
+  return preSearchText !== undefined && preFilter !== undefined && preSorting !== undefined ? (
```
peetzweg, line 69: Feels odd. That these values need to be defined. What if local storage is empty?

If these have not `undefined` initial values, we should set them directly in the `useState('')` hook instead. No extra stuff needed in line: 43

`_initSearch = _initSearch ? _initSearch : '';`

Also `PoolsRowList` could have default values defined in it's props for these if they could be undefined.

packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList/index.tsx
```diff
@@ -10,11 +20,41 @@ const FusePoolList = () => {
...
+    if (mounted.current) {
+      setPreSearchText(_initSearch);
+      setPreFilter(_initFilter);
+      setPreSorting([{ id: _initSortId, desc: initSortDesc }]);
+    }
+
+    return () => {
+      mounted.current = false;
+    };
+  }, [address]);
```
peetzweg, line 56: Why execute this useEffect on "address" change? The config in the localstorage is not address specific. 

If you only want it to be run once it's mounted it can be just `[]`.

packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList/index.tsx
smartdevcom: nvm, fixed

packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList/index.tsx
smartdevcom: Fixed

packages/ui/components/pages/Fuse/FusePoolsPage/FusePoolList/index.tsx
smartdevcom: fixed

peetzweg (approved):

peetzweg:
Can we hide the "Your Supply/Borrow" Columns on the Pool List as well as on the individual Pool pages, if no wallet is connected?

Maybe this react-table API can be used for it:
https://tanstack.com/table/v8/docs/api/features/column-visibility

Screenshot 2022-10-06 at 14 04 29


Screenshot 2022-10-06 at 14 05 03
Screenshot 2022-10-06 at 14 04 47

packages/ui/constants/index.ts
```diff
@@ -93,3 +93,8 @@ export const COLLATERAL_FACTOR_TOOLTIP =
   'Collateral factor can range from 0-90%, and represents the proportionate increase in liquidity (borrow limit) that an account receives by depositing the asset.';
 export const ADMIN_FEE_TOOLTIP =
   "The fraction of interest generated on a given asset that is routed to the asset's admin address as a fee.";
+
+export const MIDAS_POOL_SEARCH = 'midas_pool_search';
```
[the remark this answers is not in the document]
smartdevcom: fixed

Issue #766, opened 2022-10-03
Rewards Info End Timestamp Not Set Individually


peetzweg:
Reward End Timestamp is "set" for all active markets.
I have just added a end time stamp for `jCNY` but it seems to be set for all other active markets to  the same value. This should not be the case. It's probably not set for the other markets and is just not loaded correctly.

The End Timestamp should be able to be set individually for each market.

![demo](https://user-images.githubusercontent.com/839848/193547804-19b864e7-932d-4bcc-a39e-b87ad087df43.gif)

Issue #762, opened 2022-09-30
Rework `EditFlywheelModal`


peetzweg:
If we have multiple markets this modal design is not the best :D
Probably a dropdown is better here.

<img width="1404" alt="Screenshot 2022-09-30 at 15 16 53" src="https://user-images.githubusercontent.com/839848/193278202-c454af38-8dcf-42fe-aafb-c656d7e5eb1f.png">

PR #775, opened 2022-10-03
762/766 rework editFlyWheelModal

adub671:
## Description
-changed button style layout of flywheel markets to drop down.
-Fixed the bug where end time will not change if the end time has not been set
-Added notice that indicates if end time has not been set
-Added UX enhancement where if you change the selection in the drop down, and you are in 'edit' mode, it will take you out of 'edit' mode for both end time &  setting speed
-Added error handling for fundingAmount so that transactions only can send if the amount is greater than your balance or a valid number.

## Type of change

- [x] Refactoring
- [x] Bug fix (non-breaking change which fixes an issue)

## Related Issue(s)
https://github.com/Midas-Protocol/monorepo/issues/766
https://github.com/Midas-Protocol/monorepo/issues/762

Fixes


peetzweg (approved):
Works great, was able to set the end time for all of the rewards.

PR #780, opened 2022-10-04
set gaslimit bigger for mint

smartdevcom:


packages/sdk/src/modules/FundOperations.ts
```diff
@@ -55,7 +48,10 @@ export function withFundOperations<TBase extends MidasBaseConstructor>(Base: TBa
       }
       const cToken = getContract(cTokenAddress, this.artifacts.CErc20Delegate.abi, this.signer) as CErc20Delegate;
 
-      const response = (await cToken.callStatic.mint(amount)) as BigNumber;
+      const address = await this.signer.getAddress();
+      const gasLimit = await this.getProperGasLimit(amount, address);
+
+      const response = (await cToken.callStatic.mint(amount, { gasLimit })) as BigNumber;
```
peetzweg, line 54: This is just the precheck if the tx errors or not. See the `callStatic` we also need to do it in the actual `mint` call without `callStatic` further down in the file.

smartdevcom: Ah, right

peetzweg: maybe put {from:address} in here as well so it can estimate the gas properly.

PR #794, opened 2022-10-06
used router.push instead of Link

smartdevcom:


peetzweg:
Glad it's working again with the cached data.

peetzweg (approved):

PR #803, opened 2022-10-07
solved hydrate issue of next.js

smartdevcom:
`Wallet Connect Navbar` and `Midas Logo` loaded from the first loading

And flickering `Chain Select Button Group`, I couldn't reproduce it. seems like whole content was flickering as well as button group, It looks fine on my end through.


peetzweg:
It's for sure improved but not yet totally fixed. Let's go with this for now.

PR #826, opened 2022-10-10
805 research replace apexcharts with react charts

smartdevcom:
image

image

I replaced `apexchart` with `recharts` which seems to be the best chart library for React. This chart can be customized. PR is ready. cc @peetzweg 

Preview: https://preview.midascapital.xyz/137/pool/1


peetzweg:
Uh looks nice! So it's using the same color we are using for the Supply and Borrow APY columns font color?

peetzweg:
Smart to add `preview`, just wanted to do it now haha, best feature :)

smartdevcom:
> Uh looks nice! So it's using the same color we are using for the Supply and Borrow APY columns font color?

Yes, the same

peetzweg:
Can we somehow elegantly add the `%` to the x and y axis values?
As well as a axis description? X-Axis => `Utilization`  and Y-Axis => `Rate`

peetzweg:
This hover thing looks great. Maybe we can reduce the border radius and border widht a bit so it looks more like our other hover tooltips/popovers.

Screenshot 2022-10-12 at 11 46 00

carlomazzaferro:
damn this looks good!

smartdevcom:
All issues are fixed cc @peetzweg

peetzweg:
The current utilization value should be shown on the graph.


Screenshot 2022-10-12 at 15 00 11

peetzweg:
Looks good. Can we maybe cut the yaxis at 100%? So we get a slightly more resolution for the small values before the kink of the graph?
> 100% utilization should practically never happen.

smartdevcom:
image

peetzweg:
Can you maybe also prevent loading of the graph at all if the asset is protected?

demo

peetzweg:
Looks good. Can we maybe cut the yaxis at 100%? So we get a slightly more resolution for the small values before the kink of the graph?
great than 100% utilization should practically never happen.

smartdevcom:
image

like this?

smartdevcom:
This is because supply and borrow rate will be higher than 100% when utilization is 100%, so `Rate` Y Axis limit is over 100, so if we cut off the YAxis then graph should be cut off as well, is it Okay?

smartdevcom:
> Looks good. Can we maybe cut the yaxis at 100%? So we get a slightly more resolution for the small values before the kink of the graph? great than 100% utilization should practically never happen.

I only showed to 100% utilization I guess, you mean the rates?

peetzweg:
> > Looks good. Can we maybe cut the yaxis at 100%? So we get a slightly more resolution for the small values before the kink of the graph? great than 100% utilization should practically never happen.
> 
> I only showed to 100% utilization I guess, you mean the rates?

Ah you are right, I wanted to cap the wrong thing. All good! 👍 🤦

PR #829, opened 2022-10-11
sorted tvl tooltip

smartdevcom:
image

sorted by usd amount on tvl tooltip per chain

### DIFF
--- a/packages/ui/hooks/fuse/useTVL.ts
+++ b/packages/ui/hooks/fuse/useTVL.ts
@@ -12,14 +12,15 @@ export const fetchFuseNumberTVL = async (midasSdk: MidasSdk) => {
   return Number(utils.formatUnits(tvlNative, decimals));
 };
 
-interface CrossChainTVL {
-  [chainId: string]: {
+type CrossChainTVL = Map<
+  string,
+  {
     value: number;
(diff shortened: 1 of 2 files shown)


packages/ui/hooks/fuse/useTVL.ts
```diff
@@ -12,14 +12,15 @@ export const fetchFuseNumberTVL = async (midasSdk: MidasSdk) => {
   return Number(utils.formatUnits(tvlNative, decimals));
 };
 
-interface CrossChainTVL {
-  [chainId: string]: {
+type CrossChainTVL = Map<
```
peetzweg, line 15: You could also use `Record` if you like it better, it's then just a normal js object again. 🤷 
```typescript
Record<string,
{
     value: number;
     symbol: string;
     name: string;
     logo: string;
   }>
```

peetzweg (approved):

packages/ui/hooks/fuse/useTVL.ts
smartdevcom: I think Normal js object couldn't sort by values because it is sorted by keys automatically, so we should use Map. Am I missing something?

PR #831, opened 2022-10-11
enabled assets

smartdevcom:
image

image


carlomazzaferro:
Interesting, @peetzweg  mind taking a look to be sure this doesn't interfere with the rest of the SDK functionality?

packages/ui/hooks/useUSDPrices.ts
```diff
@@ -3,59 +3,34 @@ import { useQuery } from '@tanstack/react-query';
...
-  const cgIds = chainIds.map((id) => ChainIdCoingeckoIdMapping[id]);
-
-  const { data } = await axios.get(
-    `https://api.coingecko.com/api/v3/simple/price?vs_currencies=usd&ids=${cgIds.join(',')}`
-  );
-
-  return cgIds
-    .map((cgId): [string, Price] => {
-      if (data[cgId])
-        return [ChainIdCoingeckoIdMapping[cgId], { value: data[cgId].usd, symbol: '$' }];
```
peetzweg, line 41: The beauty of this `value` & `symbol` pair was to be able to show make it any currency and not hardcode it to be `usd` everyhwere.
This way we could fall back to just show stuff in `native` value as well if no `usd` price is available. 🤷

peetzweg:
Can you refactor this hook into it's own file, it's quite confusing.
I thought I have a different local file.

https://github.com/Midas-Protocol/monorepo/blob/972271ad83c82797ec668e2867be9dd423b2b683/packages/ui/hooks/fuse/useFusePools.ts#L87

peetzweg:
Also it seems we are still making request to the backend, which than makes requests on the users behalf to RPCs nodes which can get rate limited as we are not paying for these. 
I thought we refactored it to make all of these RPC requests from the frontend, so the users ip address make the requests to these RPC nodes.

https://github.com/Midas-Protocol/monorepo/blob/972271ad83c82797ec668e2867be9dd423b2b683/packages/ui/hooks/fuse/useFusePools.ts#L98-L101

smartdevcom:
> Also it seems we are still making request to the backend, which than makes requests on the users behalf to RPCs nodes which can get rate limited as we are not paying for these. I thought we refactored it to make all of these RPC requests from the frontend, so the users ip address make the requests to these RPC nodes.
> 
> https://github.com/Midas-Protocol/monorepo/blob/972271ad83c82797ec668e2867be9dd423b2b683/packages/ui/hooks/fuse/useFusePools.ts#L98-L101

I removed only `api/tvlData` endpoint. I will remove that api as well

peetzweg:
So what is doing this PR again? Enabled assets? Hasn't neon been usable all along? Or is it adding arbitrum?

carlomazzaferro:
> So what is doing this PR again? Enabled assets? Hasn't neon been usable all along? Or is it adding arbitrum?

As far as I understood, this enabled chains on development, and enabled pool creation there as well. But I see now those are already available so not entirely sure.



@smartdevcom , a brief summary of what this PR is achieving would be great

PR #841, opened 2022-10-12
Fix  bsc flywheel deployment

peetzweg:
Screenshot 2022-10-12 at 14 21 58

### DIFF
--- a/packages/sdk/src/modules/Flywheel.ts
+++ b/packages/sdk/src/modules/Flywheel.ts
@@ -221,16 +221,15 @@ export function withFlywheel<TBase extends FuseBaseConstructorWithCreateContract
         this.signer
       ) as MidasFlywheel__factory;
       const addressOfSigner = await this.signer.getAddress();
-      const mfw = await midasFlywheel.deploy();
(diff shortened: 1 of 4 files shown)


packages/sdk/src/modules/Flywheel.ts
```diff
@@ -221,16 +221,15 @@ export function withFlywheel<TBase extends FuseBaseConstructorWithCreateContract
...
-      await flywheelCoreInstanceWithSigner.initialize(
+      const flywheelCore = await midasFlywheel.deploy();
+      const initializeTx = await flywheelCore.initialize(
         rewardTokenAddress,
         options?.rewardsAddress || constants.AddressZero,
         options?.boosterAddress || constants.AddressZero,
         options?.ownerAddress || addressOfSigner
       );
-      return flywheelCoreInstance;
+      await initializeTx.wait();
```
peetzweg, line 231: just fyi @carlomazzaferro as you did the change. We missed to wait for the `initialize()` call to be successfull before continueing with the rest of the flywheel transactions. It worked on polygon as it was faster than the person clicked but that was not the case on BSC.

Issue #632, opened 2022-09-07
Use APY Api Endpoints If Available


peetzweg:
We currently have our own custom solution to calculate APYs for Plugins, however it's not working very reliably. The markets need to be interacted with commonly and the number differs from the actual strategy we are referring too. 

Most if not all of the strategies we apply offer an API to get the APY of the strategy, we should be able to use this API and only fallback to our APY.

Best example would be beefy, but not sure if we have public access to this.
https://github.com/beefyfinance/beefy-api

peetzweg:
Beefy API
Site: https://api.beefy.finance/apy
Repo: https://github.com/beefyfinance/beefy-api

peetzweg:
Ellipsis:
https://api.ellipsis.finance/api/getAPRs

PR #851, opened 2022-10-13
Track External APY if `ExternalAPYProvider` is Available

peetzweg:
Resolves https://github.com/Midas-Protocol/monorepo/issues/632

Adds an abstract class `ExternalAPYProvider`.
https://github.com/Midas-Protocol/monorepo/blob/a227674f17eb2bada9d3ff32d6c3aa965af251e6/packages/functions/src/providers/apy/ExternalAPYProvider.ts#L3-L5

Which is implemented for Beefy 
https://github.com/Midas-Protocol/monorepo/blob/a227674f17eb2bada9d3ff32d6c3aa965af251e6/packages/functions/src/providers/apy/BeefyAPYProvider.ts#L9

These if a Provider is available for a Strategy:
https://github.com/Midas-Protocol/monorepo/blob/a227674f17eb2bada9d3ff32d6c3aa965af251e6/packages/types/src/plugin.ts#L1-L7
https://github.com/Midas-Protocol/monorepo/blob/a227674f17eb2bada9d3ff32d6c3aa965af251e6/packages/functions/src/providers/apy/index.ts#L5-L9

It will than be used to fetch the data:
https://github.com/Midas-Protocol/monorepo/blob/a227674f17eb2bada9d3ff32d6c3aa965af251e6/packages/functions/src/controllers/plugins.ts#L20-L30


packages/functions/src/providers/apy/BeefyAPYProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+      await this._fetchBeefyAPYs();
+    }
+
+    const apy = this.beefyAPYs![beefyID];
+    if (!apy) throw `BeefyAPYProvider: Beefy ID: "${beefyID}" not included in Beefy APY data`;
+
+    return apy;
+  }
+
+  private async _fetchBeefyAPYs() {
```
carlomazzaferro, line 36: Probably would be a good idea to cache this call somehow -- though I believe we're already doing it with Supabase. 

Mind giving a sense of how often would this be called?

packages/chains/src/bsc/plugins.ts
```diff
@@ -1,12 +1,12 @@
...
+import { assetSymbols, DeployedPlugins, Strategy, underlying } from "@midas-capital/types";
 
 import assets from "./assets";
 
 const deployedPlugins: DeployedPlugins = {
   "0x10C90bfCFb3D2A7ae814dA1548ae3a7fC31C35A0": {
     market: "0x34ea4cbb464E6D120B081661464d4635Ca237FA7",
     name: "Bomb",
-    strategy: "BombERC4626",
+    strategy: Strategy.Bomb,
```
carlomazzaferro, line 9: Don't particularly love to add yet another property here that is quite similar to others, but I am OK with the solution for now

carlomazzaferro:
Very clean, I like how non-intrusive these changes are. In particular, it seems like a very good way to make such feature easily extensible with minimal overhead.

Approved

carlomazzaferro (approved):

packages/chains/src/bsc/plugins.ts
peetzweg: It's the same property but more strictly typed, so it we can make sure if we add new plugins we use a valid value.

packages/functions/src/providers/apy/BeefyAPYProvider.ts
peetzweg: It's just called once. As it's using the same instance of the BeefyAPYProvider. It will just call fetchBeefyAPY again if it's not available. Which should only be on the first request. If no BeefyPlugin is even available it will never be called.

Not very transparent for sure. Wasn't able to come up with a more transparent and enforced init code yet. Each APY Provider has to make sure on it's own it's to fetch as less as possible during multiple lookups.

https://github.com/Midas-Protocol/monorepo/pull/851/files#diff-6baebe0af0f07ff25c44719fa7224337f1b8b5d85847ba7a4545f6dc976ee624R26-R28

PR #852, opened 2022-10-13
UniswapV3LiquidatorFunder config for the bots

coolnft:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/chains/src/arbitrum/params.ts
```diff
@@ -10,6 +10,7 @@ const specificParams: ChainParams = {
     chainIdHex: "0xA4B1",
     name: "Arbitrum One",
     shortName: "Arbitrum",
+    uniswapV3Fee: 3000,
```
vminkov, line 13: should there be any configuration of the fee level for different assets/pairs?

packages/types/src/chain.ts
```diff
@@ -58,6 +59,7 @@ export type ChainAddresses = {
   W_TOKEN_USD_CHAINLINK_PRICE_FEED: string;
   UNISWAP_V2_ROUTER: string;
   UNISWAP_V2_FACTORY: string;
+  UNISWAP_V3_ROUTER?: string;
```
carlomazzaferro, line 62: I suggest we use something like this: https://github.com/Midas-Protocol/monorepo/blob/feature/security/packages/types/src/chain.ts

carlomazzaferro: Because each chain will also likely need as specific PAIR_INIT_HASH and quoter addresses

coolnft: we don't need `pair_init_hash` for v3 and we can get `quoter` address from deployments.

packages/chains/src/arbitrum/params.ts
coolnft: @vminkov @carlomazzaferro please check this update.

carlomazzaferro: Looks good now

vminkov (approved):

PR #879, opened 2022-10-20
Deploy/0.4.2 

vminkov:
## Description

currently deployed fully for chapel, bsc mainnet, moonbeam and almost fully (without the markets/pools upgrades) for polygon


peetzweg:
I think we should merge it into development first, we also need it there and than go to main. ✌️

carlomazzaferro (approved):

PR #881, opened 2022-10-20
split useNativeTokenBalance

smartdevcom:
Mocha test is passed in the SDK. cc @peetzweg


packages/sdk/src/modules/FundOperations.ts
```diff
@@ -80,17 +80,9 @@ export function withFundOperations<TBase extends MidasBaseConstructor>(Base: TBa
       return await this.mint(cTokenAddress, amount);
     }
 
-    async repay(cTokenAddress: string, underlyingTokenAddress: string, isRepayingMax: boolean, amount: BigNumber) {
+    async repayBorrow(cTokenAddress: string, isRepayingMax: boolean, amount: BigNumber) {
```
peetzweg, line 83: Why do we now have two repay functions?
`repayBorrow` & `repay`?

Aren't we always repaying a borrowed token?

peetzweg:
What is changed in this PR again? A bit confused.

packages/sdk/src/modules/FundOperations.ts
smartdevcom: `repay` = `approved` + `repayBorrow`. So `repay` has 2 steps for transactions, it needs to be separated to show tx to users transparently

smartdevcom:
> What is changed in this PR again? A bit confused.

As I said before, I pushed the codes which show transaction steps for supply directly to `main` with my fault, so after that I made pr and implemented for `repay`, that is PR for that

PR #898, opened 2022-10-23
showing strategy score

smartdevcom:
image


smartdevcom:
Seems like `chainConfig.chainAddresses.UNISWAP_V3` == `undefined` so UI is broken when wrapping like this `const SecurityBaseWithModules = withErc4626StrategyScorer(withChainLinkOracleScorer(withUniswapV3OracleScorer(SecurityBase)));`

So I removed them from wrapping for now cc @carlomazzaferro 

image


```
if (chainConfig.chainAddresses && chainConfig.chainAddresses.UNISWAP_V3) {
      this.quoterContract = chainConfig.chainAddresses.UNISWAP_V3.QUOTER_V2;
      this.uniV3Factory = chainConfig.chainAddresses.UNISWAP_V3.FACTORY;
      this.uniV3PairInitHash = chainConfig.chainAddresses.UNISWAP_V3.PAIR_INIT_HASH;
    } else {
      throw new Error("UniswapV3 Config not found");
    }
```

carlomazzaferro:
> Seems like `chainConfig.chainAddresses.UNISWAP_V3` == `undefined` so UI is broken with `UniswapV3 Config not found` when wrapping like this `const SecurityBaseWithModules = withErc4626StrategyScorer(withChainLinkOracleScorer(withUniswapV3OracleScorer(SecurityBase)));`
> 
> So I removed them from wrapping for now cc @carlomazzaferro
> 
> image
> 
> ```
> if (chainConfig.chainAddresses && chainConfig.chainAddresses.UNISWAP_V3) {
>       this.quoterContract = chainConfig.chainAddresses.UNISWAP_V3.QUOTER_V2;
>       this.uniV3Factory = chainConfig.chainAddresses.UNISWAP_V3.FACTORY;
>       this.uniV3PairInitHash = chainConfig.chainAddresses.UNISWAP_V3.PAIR_INIT_HASH;
>     } else {
>       throw new Error("UniswapV3 Config not found");
>     }
> ```

That is fine for now, I'll take a look to see what better resolution we can find.

Just FYI: not every chain has UniV3 deployed, so this configuration should really be optional, and everything should work if not defined

smartdevcom:
Pls review this PR cc @carlomazzaferro @peetzweg  what would be improved more and let me know if I am missing sth?

carlomazzaferro:
> Pls review this PR cc @carlomazzaferro @peetzweg what would be improved more and let me know if I am missing sth?

For now this looks fine to me, as long as we have a link to the docs. 

The next improvement will likely be implementing something more like Beefy's in which we should exactly what factors contributed to the score in the UI. 

For that, we'll need to refactor the security package to return a "detailed score" which gives a breakdown of each aspect. Let's create a ticket for that


I'll defer the UI implementation review to @peetzweg

peetzweg:
@carlomazzaferro looks quite sad to me. If we only have a single value, we should just add it to the rest of the values on the right with a tooltip.

Or we refactor the scores to get the details and be able to show them as well. Is that blocked by anything? Otherwise maybe @smartdevcom can do it?

Screenshot 2022-10-28 at 09 44 21

smartdevcom:
> @carlomazzaferro looks quite sad to me. If we only have a single value, we should just add it to the rest of the values on the right with a tooltip.
> 
> Or we refactor the scores to get the details and be able to show them as well. Is that blocked by anything? Otherwise maybe @smartdevcom can do it?
> 
> Screenshot 2022-10-28 at 09 44 21

I guess `Security` is just started and there should be more features in it to show in the UI so more info might be there in UI

carlomazzaferro:
> > @carlomazzaferro looks quite sad to me. If we only have a single value, we should just add it to the rest of the values on the right with a tooltip.
> > Or we refactor the scores to get the details and be able to show them as well. Is that blocked by anything? Otherwise maybe @smartdevcom can do it?
> > Screenshot 2022-10-28 at 09 44 21
> 
> I guess `Security` is just started and there should be more features in it to show in the UI so more info might be there in UI

I think it would make sense for @smartdevcom to get familiar with Security and implement the return of the strategy details as well. It would be good for a variety of reasons:
- Getting familiar with the security package and understanding the calculations
- Designing a solution that will play well with the FE, as in the end, that's the only target the strategy scoring is serving
- Will avoid back and forths between us

smartdevcom:
image

carlomazzaferro:
> image

Let's at least add a tooltip explaining what each of these are? Also, I am not sure if we even want to show the score number of each subsection, I'd just show the red or green triangle

smartdevcom:
image

smartdevcom:
Does `Strategy Total Score` need to be shown here? Or just put the total score number next to `safety score`?

image

smartdevcom:
https://preview.midascapital.xyz/56/pool/1

I showed all scores with tooltip

image

peetzweg:
@smartdevcom looks great! 

I find it a bit weird, that every aspect of the risk can be green but the overall score is just `0.66` and green for example. 
+ Maybe we should show the max score of `0.66/1.00` ? 
+ Add a yellow coloring, as it's seems to be either red or green.
+ show individual score in the tooltip of the aspect to investigate which value might be lower, resulting in the overall lower security value.

smartdevcom:
image

smartdevcom:
And what does it mean `Maybe we should show the max score of 0.66/1.00 ?` @peetzweg ?

smartdevcom:
image

```
export const SCORING_WEIGHTS = {
  COMPLEXITY: 0.3,
  TIME_IN_MARKET: 0.1,
  ASSET_RISK: 0.2,
  PLATFORM_RISK: 0.2,
};
```
Shouldn't  sum of all weights be `1`? @carlomazzaferro

carlomazzaferro:
> image
> 
> I showed `green` >0.8, `yellow`>0.6, `red`<0.6

I'm onboard with having this scoring criteria. Agree that a 0.7 with all green makes little sense

carlomazzaferro:
> image
> 
> ```
> export const SCORING_WEIGHTS = {
>   COMPLEXITY: 0.3,
>   TIME_IN_MARKET: 0.1,
>   ASSET_RISK: 0.2,
>   PLATFORM_RISK: 0.2,
> };
> ```
> 
> Shouldn't sum of all weights be `1`? @carlomazzaferro

Actually, that is right. Thanks for pointing this out. Let's change this to:

```
 export const SCORING_WEIGHTS = {
   COMPLEXITY: 0.3,
   TIME_IN_MARKET: 0.2,
   ASSET_RISK: 0.25,
   PLATFORM_RISK: 0.25,
 };
 ```

File: packages/security/src/enums.ts [hunk 1]
```diff
@@ -27,57 +27,147 @@ export const chainIdToConfig: { [chainId: number]: ChainConfig } = {
...
+  },
+  supplyCentralised: {
+    CENTRALIZED: {
+      title: "Few very powerful whales",
+      explanation:
+        "When the supply is concentrated in a few hands, they can greatly affect the price by selling. Whales can manipulate the price of the coin. The more people that have a vested interest over a coin, the better and more organic the price action is.",
+    },
+    DECENTRALIZED: {
+      title: "Decentralized",
+      explanation: "Decentralized",
```
carlomazzaferro, line 124: "The supply is concentrated is spread across many users instead of being concentrated in the hands of a few whales, which makes it so that price manipulations are less likely to occur. The more people that have a vested interest over a coin, the better and more organic the price action is.",

carlomazzaferro:
Added a comment on how to document the DECENTRALIZED property.  Otherwise, this looks great. I am approving but letting @peetzweg take a last pass on it.

carlomazzaferro (approved):

packages/security/src/enums.ts
smartdevcom: Added!

carlomazzaferro:
Missing change to score from 0-10, after that I think we can go live 👍
@peetzweg any further comments here?

smartdevcom:
image

smartdevcom:
> Missing change to score from 0-10, after that I think we can go live +1 @peetzweg any further comments here?

Fixed

packages/security/src/strategy/scoring.ts
```diff
@@ -21,99 +21,99 @@ export const SCORING_WEIGHTS = {
 export const complexityScore = (complexity: StrategyComplexity) => {
   switch (complexity) {
     case StrategyComplexity.LOW:
-      return 1;
+      return 10;
```
peetzweg, line 24: Why not leave it at 0-1 and just mutliply the resulting value in the frontend?

I think this would be better, if we ever decide to change it to 0-100 or something. 🤷

smartdevcom: makes sense

smartdevcom: Fixed

peetzweg:
Let's polish this up a bit more.

Screenshot 2022-11-03 at 09 30 36

Screenshot 2022-11-03 at 09 22 25

smartdevcom:
Fixed 

image

smartdevcom:
image

peetzweg (approved):

PR #905, opened 2022-10-25
setting the onchain configs for the jarvis and curve swap pools

vminkov:
## Description

setting the onchain configs for the jarvis and curve swap pools


carlomazzaferro (approved):
I think this looks good.

PR #913, opened 2022-10-27
890 externalapy provider for dotdot finance

peetzweg:
new `/api/rewards` enpoint returning all rewards for plugin even with reward token and flywheel.

Screenshot 2022-10-27 at 15 41 06

https://github.com/Midas-Protocol/monorepo/blob/03c69b2e24646afc1648ebd6c6d8e10dca818806/packages/types/src/rewards.ts#L1-L13


packages/chains/src/bsc/plugins.ts
```diff
@@ -4,21 +4,23 @@ import assets from "./assets";
 
 const deployedPlugins: DeployedPlugins = {
   // No plugin deployment file stored
+  // Can we delete this?
```
carlomazzaferro, line 7: Well, the pool is still there. Let's keep it for now

packages/types/src/plugin.ts
```diff
@@ -1,13 +1,14 @@
 export enum Strategy {
   Beefy = "BeefyERC4626",
   Arrakis = "ArrakisERC4626",
+  Mimo = "MimoERC4626",
```
carlomazzaferro, line 4: Dangerous, as we have not yet change the contract names yet

carlomazzaferro: Sounds great

carlomazzaferro (approved):

PR #914, opened 2022-10-28
added feature button

smartdevcom:
image

image


peetzweg:
The text button is a bit offcenter can you fix this?
It needs to be higher. It has a padding on top for no reason.


Screenshot 2022-10-28 at 12 37 40
Screenshot 2022-10-28 at 12 39 26


In your screenshot the text in the button is not centered, is it only locally for you?
Screenshot 2022-10-28 at 12 40 48

PR #936, opened 2022-11-02
rainbowkit

smartdevcom:
image
image
image


peetzweg:
Looks great! Need to test this thoroughly first a bit to give full review.

peetzweg:
There should be a proper gap between the buttons.

Screenshot 2022-11-08 at 10 31 55

peetzweg:
Data after transactions is not properly refreshed:
Not only collateral, also supplying, withdrawing etc.

brokencollateral

peetzweg (approved):

PR #966, opened 2022-11-05
improved loading skeleton

smartdevcom:
loading1


peetzweg:
Looks good! Although, why is it initially so small, basically just showing pool details? THe Asset box is very thin. In the current version it's already a bigger size.

Can we improve this? Goal would be to have the least amount of jumping elements on the page.

peetzweg:
So in the beginning it's briefly very small visible. Can't this already be a bit bigger? The Pool Details box is already big and shown at the top and than pushed down by the actual content of the page.

### T0
Screenshot 2022-11-07 at 12 05 19


### T1
Screenshot 2022-11-07 at 12 05 21

peetzweg:
So many changes, I though it would only touch the skeleton haha. Bit scared to push it to production today. So maybe tomorrow :S

smartdevcom:
> So many changes, I though it would only touch the skeleton haha. Bit scared to push it to production today. So maybe tomorrow :S

Another simple PR is here https://github.com/Midas-Protocol/monorepo/pull/975

peetzweg:
We chose this implementation:
https://github.com/Midas-Protocol/monorepo/pull/975

PR #996, opened 2022-11-11
several ui tweaks

peetzweg:
### Before

Screenshot 2022-11-11 at 13 53 08

### After

Screenshot 2022-11-11 at 13 53 09

### Before

Screenshot 2022-11-11 at 13 52 00 1

### After

Screenshot 2022-11-11 at 13 52 00


packages/ui/theme/components/Button.ts
```diff
@@ -49,12 +50,14 @@ export const ButtonStyleConfig: ComponentStyleConfig = {
...
       _hover: {
-        bg: mode('ecru80', 'ecru80')(props),
+        textDecoration: 'unset',
+        bg: mode('ecru', 'ecru')(props),
         color: mode('raisinBlack', 'raisinBlack')(props),
       },
     }),
     filter: (props) => ({
       ...theme.components.Button.variants?.solid,
+      height: '52px',
```
carlomazzaferro, line 60: Not going to claim that I understand how any of these work, but the output looks good :)

PR #1019, opened 2022-11-16
removed unused component & fixed flickering rewards token icon when s…

smartdevcom:
Peek 2022-11-17 00-48
Peek 2022-11-17 00-49


peetzweg:
Can you fix the zIndex of the token tooltip as well?

Btw have you checked if this new 9999 z index of the navbar is not overlaying Toasts?

smartdevcom:
image
 cc @peetzweg

smartdevcom:
> Can you fix the zIndex of the token tooltip as well?
> 
> Btw have you checked if this new 9999 z index of the navbar is not overlaying Toasts?
> 
> Would it be easy to just move all toast to the bottom right? That would be nice. It's overlayed by metamask anyways always. @smartdevcom

token tooltip works

image

peetzweg (approved):

Issue #1001, opened 2022-11-14
FEATURE: Deprecate Markets


peetzweg:
As recently, we paused minting and borrowing of a market for safety reasons. In the ui we treated it as a `deprecated` market as the user was unable to supply or borrow in that market anymore. However if we are doing it for safety reasons, deprecated is not the best naming for this. So we changed it to paused. However, there are some markets we want to mark as deprecated. We need a feature for this. We could either do it again via environment variables or do it via a list in supabase? This would not require a deploy of the frontend but would be more complex.

Let's go with the env version for now as the list is not very long. Or just hardcode it in the backend?

PR #1042, opened 2022-11-22
hardcoded deprecated assets in sdk

smartdevcom:
I set hardcoded deprecated assets in SDK. I guess handling this in SDK is better and simpler than in UI because we have some other calculations from assets of SDK.


peetzweg:
`market` instead of `asset` would be a better name.

`asset` is the underlying `asset` of a `market`.
So `USDC` for example could be the `underlying` of a `USDC market`, so the underlying `token` becomes a `cToken` in the `comptroller`
We can't deprecate a `USDC` we can deprecate however the `USDC Market of a Comptroller` which is a `cToken`, so `market` is a better fit for the naming here than `asset`.

We might need to `deprecate` `assets` later, but this is actually would be just removing the `asset` from `assets.ts` file to not be able to deploy any new `markets` with this underlying `asset`.

https://github.com/Midas-Protocol/monorepo/blob/65f4319015e1b9b1a3b2d914276555adf2fc027e/packages/chains/src/polygon/assets.ts#L100-L107

smartdevcom:
I am not sure if this PR is still valid because related ticket is closed as completed here https://github.com/Midas-Protocol/monorepo/issues/1001 cc @carlomazzaferro

carlomazzaferro:
Closing this, as we can deprecate markets at the contracts level

PR #1061, opened 2022-11-27
fixed console warning, fixed styles of amount modal

smartdevcom:
image
image


smartdevcom:
It seems like `getMaxBorrow` function is considered by `Liquidity` as well so it returns only borrowable amount

```
if (mode === FundOperationMode.BORROW) {
    const maxBorrow = (await midasSdk.contracts.FusePoolLensSecondary.callStatic.getMaxBorrow(
      address,
      asset.cToken
    )) as BigNumber;

    if (!maxBorrow) {
      throw new Error('Could not fetch your max borrow amount! Code: ');
    }
    return maxBorrow;
  }
```

smartdevcom:
image

breaking Amount Modal by `long token name` is fixed in this PR

peetzweg (approved):

PR #1083, opened 2022-12-05
enabled hiding markets which user doesn't supply or borrow from

smartdevcom:
Peek 2022-12-06 01-49


peetzweg:
Is this saved in between user session in local storage?

smartdevcom:
Now it's saved in localstorage

PR #1084, opened 2022-12-06
added user stats in markets list

smartdevcom:
image

Let's see how `Supply APY` and `Borrow APY` should be calculated.


smartdevcom:
image
image

peetzweg:
Need to test this on a fork, will do so probably tomorrow. We are at risk here of showing wrong data to the user, or at least confusing data. So we should add this with caution.

peetzweg:
Reviewing this now :)

peetzweg:
Can you add such information on hover for the user stat fields as well? This would be helpful for users to understand how the values are aggregated.

Screenshot 2023-01-09 at 11 26 11

---

We need to update the skeletons for the new UserStats

Screenshot 2023-01-09 at 11 25 38

peetzweg:
1) Can you move the Collateral Ratio bar below the User Stats. So only outside the box are "total pool stats".

I saw the code, maybe you can refactor the `MarketList` to be just the Filter + the Rows of markets. And the stuff above, like now the `CollateralRatio` and `UserStats` are organized on the `PoolPage/index.ts` file.


Screenshot 2023-01-16 at 12 06 03


2) While you are at it, please make all of these spacing the same. If it's all on the `PoolPage/index.ts`, it's the easiest to maintain to use the `gap` prop for the parent container of these elements.


Screenshot 2023-01-16 at 12 06 42


3) Show an estimate USD value of the supply rewards and costs inside the UserStats. That would be nice for users to see and maybe cause more deposits.
It should be in a more subtle color as these values are only "estimates" as it's depending on the prices, apys etc. .

Screenshot 2023-01-16 at 12 08 09

peetzweg (approved):

PR #1103, opened 2022-12-09
1092 show transaction summary

smartdevcom:
Peek 2022-12-10 22-08


peetzweg:
Looks great in the gif! Thanks for sharing that. Will give it a try myself.

Can we maybe remove, the "x" at the top, so user don't close it? And add a Button with a `Close` label at the bottom once all tx are finished?

smartdevcom:
> Looks great in the gif! Thanks for sharing that. Will give it a try myself.
> 
> Can we maybe remove, the "x" at the top, so user don't close it? And add a Button with a `Close` label at the bottom once all tx are finished?

In this case, user should wait for all txs finished, is it Ok?

peetzweg:
The TX link could be shown as soon as the the user presses  'confirm' in metamask. The `tx.wait()` should also ahve the tx hash which we could direclty link to the scanner. So they could observe the success/failure/progress of the transaction also in the scanner. 
So let's show the link to the scanner directly and not just after the tx has succeeded or failed.
Screenshot 2022-12-15 at 15 55 06

smartdevcom:
> The TX link could be shown as soon as the the user presses 'confirm' in metamask. The `tx.wait()` should also ahve the tx hash which we could direclty link to the scanner. So they could observe the success/failure/progress of the transaction also in the scanner. So let's show the link to the scanner directly and not just after the tx has succeeded or failed. Screenshot 2022-12-15 at 15 55 06

Ah, that makes sense

smartdevcom:
Peek 2022-12-16 09-36

Now tx scanner will be appeared before getting tx result. And also `x` button will be disappeared while acting

PR #1117, opened 2022-12-13
task to replace the broken flywheel

vminkov:
## Description

task to replace the broken flywheel


packages/sdk/tasks/one-time/loopless-booster.ts
```diff
@@ -34,3 +38,104 @@ task("loopless-booster", "deploy and a loopless booster for a flywheel")
...
+        from: deployer.address,
+        log: true,
+        args: [
+          replacingFlywheel.address, // flywheel
+          deployer.address, // owner
+          constants.AddressZero, // Authority
+        ],
+      });
+
+      // this will transfer all the rewards from the old FlywheelStaticRewards contract
```
carlomazzaferro, line 107: would add:

```
    console.log(replacingRewards.transactionHash);
    if (replacingRewards.transactionHash) await ethers.provider.waitForTransaction(replacingRewards.transactionHash);
```

carlomazzaferro: To avoid nonce re-used etc

PR #1119, opened 2022-12-14
Add collateral modal

smartdevcom:
Peek 2022-12-14 15-17


peetzweg:
Maybe we should show the whole `StatsColumn` component. As disabling a collateral can change a lot with the current users position. The borrow limit decrease and the user get's closer to liquidation range.

I know the `StatsColumn` component is not yet the best visually but we will improve it probably over time.

peetzweg:
Otherwise looks great!

peetzweg:
Do the edge cases work, whereas the user can't disable the collateral because he than has to little collateral left?

smartdevcom:
> Do the edge cases work, whereas the user can't disable the collateral because he than has to little collateral left?

It will be failed. cause it's failed from sdk first. So the modal shows `Failed`

packages/sdk/src/modules/FundOperations.ts
```diff
@@ -29,32 +30,39 @@ export function withFundOperations<TBase extends MidasBaseConstructor>(Base: TBa
...
     async approve(cTokenAddress: string, underlyingTokenAddress: string, amount: BigNumber) {
-      const token = getContract(underlyingTokenAddress, this.artifacts.EIP20Interface.abi, this.signer);
+      const token = getContract(
+        underlyingTokenAddress,
+        this.artifacts.EIP20Interface.abi,
+        this.signer
+      ) as EIP20Interface;
       const currentSignerAddress = await this.signer.getAddress();
 
       const hasApprovedEnough = (await token.callStatic.allowance(currentSignerAddress, cTokenAddress)).gte(amount);
```
peetzweg, line 40: I always wanted to move this check from the sdk into the UI. It think it's ugly to check here. Would clean this up a lot.

packages/sdk/src/modules/FundOperations.ts
```diff
@@ -76,7 +84,10 @@ export function withFundOperations<TBase extends MidasBaseConstructor>(Base: TBa
       amount: BigNumber
     ) {
       await this.approve(cTokenAddress, underlyingTokenAddress, amount);
-      await this.enterMarkets(cTokenAddress, comptrollerAddress, enableAsCollateral);
+      if (enableAsCollateral) {
```
peetzweg, line 87: Same with this. Can we move this into the UI? As it's a field set by the UI itself anyways.
`enableAsCollateral`

So only call `enterMarket` if `enableAsCollateral` is set. 

Same with `approve` only call approve if not enough already approved. 

The functions themselves should not check if they should do it or not. Cleaning these up and they will always return a `Promise<Transaction>` and not sometimes null or such stuff.

packages/sdk/src/modules/FusePools.ts
```diff
@@ -77,6 +77,7 @@ export function withFusePools<TBase extends MidasBaseConstructor>(Base: TBase) {
 
         if (_asset) {
           asset.underlyingSymbol = _asset.symbol;
+          asset.logoUrl = "https://d1912tcoux65lj.cloudfront.net/token/96x96/" + _asset.symbol.toLowerCase() + ".png";
```
peetzweg, line 80: What is this one about? A bit confused why we set this. Only thing which comes to mind is the strategy logo but we have that already o.O?

peetzweg:
Although all these changes are good, it's touching a lot more stuff than just adding a Collateral Modal. This makes this more work to review and properly test the stuff. 
Some critical actions are touched here. So feels unreasonable to merge this without time to test it on a fork and having backup time to fix stuff in fit breaks in production.

peetzweg:
I have limited availability right now and can test it the earliest tomorrow on fork.

smartdevcom:
Lots of changes come from just `transaction summary` PR because I cloned it from it and updated according to the latest changes of that branch. pls check commits

packages/sdk/src/modules/FusePools.ts
smartdevcom: I think we could set the token info in the SDK directly rather than in UI because we need to call hook `useTokenData` in UI to only get token image

packages/sdk/src/modules/FundOperations.ts
smartdevcom: Makes sense! I will do like that

carlomazzaferro:
UI completely freezes, see recording: https://drive.google.com/file/d/12grFtqD3QFU8A6QHsRcQOswGwoSWaAJp/view?usp=share_link

smartdevcom:
This seems be because UI is calling contracts multiple times and some hooks are not cached as I mentioned before(for `rewards` hooks) https://discord.com/channels/930077813585899561/1052282231630090250/1053294537453346977 cc @peetzweg . Couldn't we cache all hooks? cc @peetzweg

smartdevcom:
> This seems be because UI is calling contracts multiple times and some hooks are not cached as I mentioned before(for `rewards` hooks) https://discord.com/channels/930077813585899561/1052282231630090250/1053294537453346977 cc @peetzweg . Couldn't we cache all hooks? cc @peetzweg

PR is ready for this here https://github.com/Midas-Protocol/monorepo/pull/1145

carlomazzaferro (approved):

PR #1120, opened 2022-12-14
showed end date of flywheel

smartdevcom:
image


peetzweg (approved):
Nice. I've made some minor improvements. Thought this would be uglier to implement. Will merge maybe tomorrow before the holidays if we have enough time to catch eventual bugs.

PR #1121, opened 2022-12-14
SDK Pruning

peetzweg:
Cleans up the imports and distribution of types, abis and artifacts.

Deletes a bunch of unused Artifacts (huge files), imports `typechain` stuff nicer and generates `ABI` files from the `artifacts` using `abimate` a package, I wrote.
It extracts the abi form the artifact and create a typescript file of it, which can be used with `ethers` but at best with `wagmi/core` stuff. With `wagmi/core` we can only use single functions of an abi in a typesafe way.

https://github.com/peetzweg/abimate#usage

New import paths

+ `@abis` 
+ `@typechain`
+ `@artifacts`


packages/sdk/chainDeploy/helpers/dynamicFlywheels.ts
```diff
@@ -1,7 +1,7 @@
 import { constants } from "ethers";
 
 import { FuseFlywheelDeployFnParams } from "..";
-import { MidasFlywheel } from "../../lib/contracts/typechain/MidasFlywheel";
+import { MidasFlywheel } from "@typechain/MidasFlywheel";
```
carlomazzaferro, line 4: that is _nice_

packages/sdk/src/Artifacts.ts
```diff
@@ -1,177 +1,71 @@
...
   UniswapTwapPriceOracleV2Root,
-  Unitroller,
   WhitePaperInterestRateModel,
-  MidasERC4626,
 };
 
+export type Artifacts = typeof ARTIFACTS;
 export default ARTIFACTS;
 
 export const oracleConfig = (deployments: ChainDeployment, artifacts: Artifacts, availableOracles: Array<string>) => {
```
carlomazzaferro, line 71: It is possible that we might be able to remove this oracle config here

packages/sdk/src/MidasSdk/abi/ERC20.json
```diff
@@ -1,222 +0,0 @@
-[
```
carlomazzaferro, line 1: We actually do use this one in many places in the bots and such

carlomazzaferro: E.g. https://github.com/Midas-Protocol/monorepo/blob/fix/broken-flywheel-replace/packages/bots/liquidator/src/utils/approveTokensToSafeLiquidator.ts#L2

packages/sdk/src/index.ts
```diff
@@ -1,5 +1,5 @@
 export { default as MidasSdk } from "./MidasSdk";
-export { default as ERC20Abi } from "./MidasSdk/abi/ERC20.json";
+export { default as ERC20Abi } from "@abis/EIP20Interface";
```
carlomazzaferro, line 2: Okay, I see now how we can use it

peetzweg:
Unfortunately I had to get rid of the typescript paths again as they are not resolved during compilation. 🙃
It's by design. Not sure how to fix this.

Overall the frontend bundle got smaller as well. Not yet ideal though. GUess once we get rid of the rest of the artifacts.

### before
Screenshot 2022-12-19 at 16 00 08

### after
Screenshot 2022-12-19 at 15 58 20

smartdevcom (approved):

PR #1138, opened 2022-12-21
Fix/fpd indexes

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/monitors/liquidity/src/services/pool.ts
```diff
@@ -18,7 +18,7 @@ export class PoolService {
 
   async getPoolsWithAsset(): Promise<Comptroller[]> {
     const poolsWithAsset: Comptroller[] = [];
-    const pools = await this.sdk.contracts.FusePoolDirectory.getAllPools();
+    const [, pools] = await this.sdk.contracts.FusePoolDirectory.getActivePools();
```
peetzweg, line 21: So pool id is irrelevant here?

carlomazzaferro: Why would we need it here?

PR #1180, opened 2023-01-10
moved collateral to additional panel

smartdevcom:
Peek 2023-01-11 00-58


peetzweg:
Super cool, love the gifs as well. 👍

packages/ui/components/pages/PoolPage/MarketsList/CollateralModal/PendingTransaction.tsx
```diff
@@ -33,7 +33,7 @@ export const PendingTransaction = ({
             All Done!
           </Text>
           <Text variant="mdText" fontWeight="bold">
-            You {asset.membership ? 'disabled' : 'enabled'} {asset.underlyingSymbol} as collateral
```
peetzweg, line 36: Oh this was wrong? Good catch.

peetzweg:
Maybe the `Collateral.tsx` file should be moved as it no longer belongs to the MarketList components?
Not sure where we keep the Funding Operation Components or if there are any separate ones.

peetzweg (approved):

smartdevcom:
Rearranged components inside `MarketList` component

PR #1188, opened 2023-01-11
fixed text size

smartdevcom:
image


smartdevcom:
And also I fixed repay max value in case of native token needs to be wrapped.

Before, the balance of native token was set as max value so maybe `0.1600...` BNB is your wallet native token balance in posted your gif. It needs to be compared to max value. Check PR cc @peetzweg

peetzweg (approved):

PR #1190, opened 2023-01-12
added tx steps for single claim

smartdevcom:
ezgif com-gif-maker

ezgif com-gif-maker


packages/ui/components/pages/Fuse/Modals/ClaimRewardsModal/PendingTransaction.tsx
```diff
@@ -0,0 +1,129 @@
...
+  failedStep: number;
+  steps: TxStep[];
+  isClaiming: boolean;
+  poolChainId: number;
+  assetPerRewardToken: { [rewardToken: string]: SupportedAsset | undefined };
+}) => {
+  const successToast = useSuccessToast();
+  const errorToast = useErrorToast();
+
+  const addToken = async (asset: SupportedAsset) => {
```
peetzweg, line 30: This would be a great hook!

I guess we have something similar in the other PendingTransaction components?


Make sure to return a `useCallback()` as it's more efficient. Without it every rerender, will create this function a new.

```typescript
const useAddTokenToWallet = ({underlying,symbol})=>useCallback(()=>{
// do stuff here
},[underlying,symbol])
```

peetzweg:
Super quick! The last claim of main failed. Which is okay and can happen. Can you however make sure that the error is not bloating the toast? The toast was so big, I can't click the `x` to close it.

Screenshot 2023-01-12 at 13 54 23

smartdevcom:
Set toast max height with scroll

image

packages/ui/components/pages/Fuse/Modals/ClaimRewardsModal/PendingTransaction.tsx
smartdevcom: Created `useAddTokenToWallet` hook

peetzweg:
> Set toast max height with scroll
> 
> image

That was a good idea 👍

peetzweg (approved):

PR #1191, opened 2023-01-12
Feature: Borrow Caps

smartdevcom:
image
image

Now admin could set borrow caps for each collateral asset


smartdevcom:
image

peetzweg:
https://github.com/Midas-Protocol/monorepo/pull/1191#issuecomment-1385073950

@vminkov seeing some ui like this, I'm getting a bit confused. Can you give us some insight how it is determined on the Smart Contract side which collateral is used where? Maybe there is another asset I want my BTC to be used as collateral as well.

Can you point us to the functions in the contracts which calculate this stuff?

vminkov:
> 

it is calculated here https://github.com/Midas-Protocol/contracts/blob/development/contracts/compound/Comptroller.sol#L874

peetzweg (approved):

PR #1197, opened 2023-01-13
System admin roles

vminkov:
## Description

splitting the deployer into different roles


packages/sdk/hardhat.config.ts
```diff
@@ -71,10 +71,15 @@ const config: HardhatUserConfig = {
   },
 
   namedAccounts: {
-    deployer: { default: 0 },
-    alice: { default: 1 },
-    bob: { default: 2 },
-    rando: { default: 3 },
+    upgradesAdmin: { default: 0 },
```
carlomazzaferro, line 74: Why are they kept at zero?

vminkov: In the beginning it is fine to start with all named accounts pointing to the same address - just to test things work. Then we can change these to different ids of the same mnemonic and run the `system:admins:split` task to actually split the roles. After that the roles can be transferred to addresses of different mnemonics.

carlomazzaferro:
Do we have a document with the proposed separation of roles as outlined here? Think having it in a plain doc / notion would help, though I think you already had done this

vminkov:
> Do we have a document with the proposed separation of roles as outlined here? Think having it in a plain doc / notion would help, though I think you already had done this

here https://www.notion.so/System-admin-roles-ddaae59b5a2a41e7a257fd390b91cf7e

Issue #1199, opened 2023-01-15
Validation of live asset configuration


carlomazzaferro:
Design a tool to fetch all deployed assets, and validate what configuration they have:
- Which oracle, and if it is deployed
- Which liquidation strategy
Tool should also tell whether we have assets in our `chains` package that is not live, and suggest to set it to `disabled`

NikMidas:
Per Carlo, this is complete

PR #1200, opened 2023-01-15
refactor chainIdToConfig

carlomazzaferro:
1. Removes the  multiple `chainIdToConfig` definitions from everywhere and places it where it makes the most sense, in the `chains` packages
2. https://github.com/Midas-Protocol/monorepo/pull/1200/files#diff-2d38283ff0f62274241334f6d423956101af87a0b0d5c34661f04521ab97297e is the only place where a somewhat larger refactor was needed, as this file assumed that all the exports from the `chains` packages were just the chain configs themselves
3. Adds a new set of hardhat tasks (WIP) that aims at aggregating all the available / deployed assets and their configuration, aimed at making it easier to see which configs are discrepant or missing. See https://github.com/Midas-Protocol/monorepo/issues/1199


peetzweg:
Thanks for taking the effort cleaning this stuff up :)

peetzweg (approved):

PR #1201, opened 2023-01-16
Showed more correctly for paused strategies

smartdevcom:


packages/types/src/rewards.ts
```diff
@@ -1,5 +1,6 @@
 export interface AbstractReward {
   apy?: number;
+  status?: string;
```
peetzweg, line 3: Let's type this stronger with either a own `enum` or at least const strings like

```
status?: 'paused'|'unknown';
```

Let's also add `'unknown'`. It can be useful for cases where we don't get information about apys. We still however, should make the discord notification so we know somehtings is off. And show in the frontend.

`Information about the rewards of this plugin are currently not available to us. We are are trying to resolve this as soon as possible`

https://github.com/Midas-Protocol/monorepo/blob/effd8623ddd97d4b8ab7af0e933eec043bc8af8c/packages/functions/src/providers/rewards/plugins/BeefyAPYProvider.ts#L46-L52

peetzweg (approved):

PR #1203, opened 2023-01-17
added addresses link for oracle and irm

smartdevcom:
image


peetzweg:
IRM looks good. Oracle is always returning the `MasterPriceOracle` (MPO) which is just forwarding the `price()` requests to the proper `Oracle` of the underlying.

We need to fetch the actual `oracle` address from the `MasterPriceOracle` by reading the address of the `mapping(address => IPriceOracle) public oracles;` of the `MasterPriceOracle` contract.

So you need to create an instance of the MPO with maybe a basic harcoded ABI which looks like the following and call the `oracles(underlyingAddress)` function.


```JSON
[
  {
      "inputs": [
        {
          "internalType": "address",
          "name": "",
          "type": "address"
        }
      ],
      "name": "oracles",
      "outputs": [
        {
          "internalType": "contract IPriceOracle",
          "name": "",
          "type": "address"
        }
      ],
      "stateMutability": "view",
      "type": "function"
    }
]
    ```

**Comment by smartdevcom** [comment]
Fixed and I used `sdk.createMasterPriceOracle(sdk.provider)` instead of hardcoded abi

File: packages/ui/hooks/fuse/useOracle.ts [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+  return useQuery(
+    ['useOracle', comptrollerAddress, sdk?.chainId],
+    async () => {
+      if (comptrollerAddress && sdk) {
+        const comptroller = sdk.getComptrollerInstance(comptrollerAddress);
+
+        const mpoAddress = await comptroller.oracle();
+
+        const mpo = sdk.createMasterPriceOracle(sdk.provider);
+        const oracle = await mpo.oracles(mpoAddress);
```
carlomazzaferro, line 17: This should actually be `mpo.oracles(underlyingAddress)`

smartdevcom: Fixed!

carlomazzaferro (approved):

PR #1210, opened 2023-01-18
analyzed bundle size

smartdevcom:
Added package and tried to run `yarn analyze:ui`

results:

client.html
image

nodejs.html
image


smartdevcom:
As you can see `nodejs.html`, `tokenData.ts` call is so big but I guess we don't need this call because we only need this  call for assets which isn't defined in our SDK but is this still the case?

I think we already get name, symbol, decimals from assets defined in SDK.

peetzweg (approved):

PR #1213, opened 2023-01-18
992 reduce deployments json file size

peetzweg:
Reduced bundle size even further to speed up loading speed. Nothing inherently big left as far as I can see.

Screenshot 2023-01-18 at 16 47 36


peetzweg:
@carlomazzaferro can you give me a hint in the right direction here how to fix this?
I removed all the abis from the deployments files. ABIs are now only available from the sdk package.
Screenshot 2023-01-18 at 17 17 23

carlomazzaferro:
> @carlomazzaferro can you give me a hint in the right direction here how to fix this? I removed all the abis from the deployments files. ABIs are now only available from the sdk package. Screenshot 2023-01-18 at 17 17 23

1. https://github.com/Midas-Protocol/monorepo/blob/development/packages/monitors/oracle/src/services/admin.ts#L2-L3
2. https://github.com/Midas-Protocol/monorepo/blob/development/packages/monitors/oracle/src/services/pool.ts#L2

PR #1231, opened 2023-01-25
hardcode market names

smartdevcom:


packages/ui/hooks/usePoolInfo.ts
```diff
@@ -0,0 +1,36 @@
...
+import axios from 'axios';
+
+export type usePoolInfoData = {
+  markets: {
+    [marketAddress: string]: {
+      name?: string;
+    };
+  };
+};
+export function usePoolInfo(chainId?: string, comptrollerAddress?: string) {
```
peetzweg, line 11: Maybe we should start calling these hooks `useAPIComptroller`

We have a lot of hooks and it's confusing which does what. Quite sure we have some hooks which are called PoolData or so as well.

So let's start using names likes `useAPI` for our api related hooks.

packages/ui/pages/api/chain/[chainId]/comptroller/[comptrollerAddress].ts
```diff
@@ -0,0 +1,38 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
+import * as yup from 'yup';
+
+import { ALL_POOLS_INFO, SUPPORTED_NETWORKS_REGEX, VALID_ADDRESS_REGEX } from '@ui/constants/index';
+
+export type ResponseType = {
```
Comment (peetzweg) [inline packages/ui/pages/api/chain/[chainId]/comptroller/[comptrollerAddress].ts:6]: Lets call it `GetComptrollerResponse`

As it's all in the same nextjs project, you should be able to use this type in the UI code as well.

peetzweg:
@carlomazzaferro are there any live markets you want to rename yet?

You mentioned the usdc market in the stall pool on moonbeam. Should we rename it to 
`USDC (non autocompounding)`?

carlomazzaferro:
Not yet! Will keep you posted

packages/ui/hooks/usePoolInfo.ts
smartdevcom: Fixed

packages/ui/pages/api/chain/[chainId]/comptroller/[comptrollerAddress].ts
Comment (smartdevcom) [reply packages/ui/pages/api/chain/[chainId]/comptroller/[comptrollerAddress].ts]: Fixed

PR #1253, opened 2023-01-30
Replace `LogRocket` with `Sentry`

smartdevcom:
`SENTRY_DSN` should be added in github env. It is gathered from here https://sentry.io/settings/midas-capital-xyz/projects/javascript-nextjs/keys/


packages/ui/pages/_app.tsx
```diff
@@ -29,6 +29,14 @@ function MidasDapp({ Component, pageProps }: AppProps) {
             <MultiMidasProvider>
               <Layout>
                 <Component {...pageProps} />
+                <button
```
peetzweg, line 32: I guess this is not needed anymore

smartdevcom: Oh, sorry

peetzweg:
Two cool features we need to debug errors submitted to sentry are `context` and `transactions`.



With `context` we should add information like:

+ is user connected?
+ to which chain is he connected

https://docs.sentry.io/platforms/javascript/enriching-events/context/

Maybe we can use the identify user feature to track the address of the user who is connected to the app:
https://docs.sentry.io/platforms/javascript/enriching-events/identify-user/


We have this genericError Handler which is nice but sometimes we know more about the error and should report the error with more information about it. You removed a bunch of logrocket exception reporting lines. Some of the were "Analytics" some where "Error" reporting.
Sentry should only be used for "Error" reporting. Let's add some function to report errors more explicitly maybe using this transaction feature from sentry:

https://docs.sentry.io/platforms/javascript/enriching-events/transaction-name/

So this way we can add information we need to debug:

i.e. Depositing

+ Token + Token Address
+ chain id
+ comptroller address

smartdevcom:
> Two cool features we need to debug errors submitted to sentry are `context` and `transactions`.
> 
> With `context` we should add information like:
> 
> * is user connected?
> * to which chain is he connected
> 
> https://docs.sentry.io/platforms/javascript/enriching-events/context/
> 
> Maybe we can use the identify user feature to track the address of the user who is connected to the app: https://docs.sentry.io/platforms/javascript/enriching-events/identify-user/
> 
> We have this genericError Handler which is nice but sometimes we know more about the error and should report the error with more information about it. You removed a bunch of logrocket exception reporting lines. Some of the were "Analytics" some where "Error" reporting. Sentry should only be used for "Error" reporting. Let's add some function to report errors more explicitly maybe using this transaction feature from sentry:
> 
> https://docs.sentry.io/platforms/javascript/enriching-events/transaction-name/
> 
> So this way we can add information we need to debug:
> 
> i.e. Depositing
> 
> * Token + Token Address
> * chain id
> * comptroller address

Added details sentry info on every tx failed like described above

smartdevcom:
image

packages/ui/utils/errorHandling.ts
```diff
@@ -1,33 +1,48 @@
...
+  error,
+  toast,
+  sentryInfo,
+}: {
+  error: any;
+  toast?: CreateToastFnReturn;
+  sentryInfo: { contextName: string; properties: { [key: string]: any } };
+}) => {
+  console.error(error);
+  console.error(JSON.stringify(error));
```
peetzweg, line 15:
```suggestion
  console.error("Raw Error", error);
```

packages/ui/utils/errorHandling.ts
```diff
@@ -1,33 +1,48 @@
...
-  console.error(e);
-  console.error(JSON.stringify(e));
+export const handleGenericError = ({
+  error,
+  toast,
+  sentryInfo,
+}: {
+  error: any;
+  toast?: CreateToastFnReturn;
+  sentryInfo: { contextName: string; properties: { [key: string]: any } };
```
peetzweg, line 12: Can we get the actual type from the sentry package somehow instead of defining it ourselves?

packages/ui/hooks/useRewards.ts
```diff
@@ -32,7 +32,6 @@ export function useRewards({ poolId, chainId }: UseRewardsProps) {
               return [];
             }),
             sdk.getFlywheelMarketRewardsByPool(poolData.comptroller).catch((exception) => {
-              // TODO LogRocket, this should never happen.
```
peetzweg, line 35: TODO means we should add logging to logrocket, now sentry here. So we know when stuff goes wrong.

peetzweg:
Thanks for adding the context and properties to the errors. ✌️

packages/ui/utils/errorHandling.ts
smartdevcom: Seems like they don't have specific type for context

smartdevcom: image

packages/ui/hooks/useRewards.ts
smartdevcom: Fixed

packages/ui/utils/errorHandling.ts
smartdevcom: Fixed

packages/ui/context/MultiMidasContext.tsx
```diff
@@ -121,15 +121,10 @@ export const MultiMidasProvider = ({ children }: MultiMidasProviderProps = { chi
   }, [signer, sdks]);
 
   useEffect(() => {
-    if (window.location.hostname === 'app.midascapital.xyz') {
```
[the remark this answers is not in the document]
smartdevcom: I have used `env` value `IS_SENTRY_ENABLED` for it

peetzweg:
@smartdevcom can you resolve the final change requests and conflics. Happy to move this to production to test it out.

PR #1267, opened 2023-02-02
showed msg only, extracted `useMaxBorrowAmount` hook from `fetchMaxAm…

smartdevcom:
image


smartdevcom:
Fixed for borrow like supply, extracted `useMaxBorrowAmount` hook from `fetchMaxAmount` func

PR #1301, opened 2023-02-13
added flag to deploy fw in UI

smartdevcom:


packages/ui/.env.dist
```diff
@@ -20,3 +20,4 @@ SUPABASE_ASSET_APY_TABLE_NAME=asset-apy-development
 
 SENTRY_DSN=
 IS_SENTRY_ENABLED=
+FEATURE_DEPLOY_FLYWHEEL=
```
peetzweg, line 23: This value is used in the frontend part of the code. It won't get this env variables as these are just available on netlify. You can test this locally. Try to set it. It should not be available in your localhost version either.

We need to prefix it with `NEXT_PUBLIC`
https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

smartdevcom: So should it be available in production only not in local?

carlomazzaferro (approved):

PR #1303, opened 2023-02-20
no need to call sdk func again

smartdevcom:
`useFusePoolData` hook calls `fetchFusePoolData` func of sdk. `useCrossFusePools` hook calls `fetchPoolsManual` func which uses `fetchFusePoolData` with pool ids so we don't need to call `fetchFusePoolData` func again but just need to filter the results of `useCrossFusePools`


peetzweg:
Happy to remove duplicate code. Just need to make 100% sure it works.

peetzweg (approved):

PR #1304, opened 2023-02-21
showed all assets user funded / Account Page

smartdevcom:
image
image
image


peetzweg:
Can you make the font size and vertical spacing similar to the pool heading?
As of know it looks a bit unrefined.


Screenshot 2023-02-24 at 10 24 23


Screenshot 2023-02-24 at 10 24 39

peetzweg:
The APYs don't seem to match up.
Screenshot 2023-02-24 at 10 30 39

peetzweg:
Account page shows "CollateralRatioBar" in loading Skeleton. We should remove it from the loading skeleton.

Screenshot 2023-02-24 at 10 27 52
Screenshot 2023-02-24 at 10 32 08

peetzweg:
Effective APY seems off, it's with the Midas Dev account. Could be due to 'preview' supabase database. But then the apys should be also off in the individual pools. 🤔 
Which pool has the broken apys on preview? 🤔 

Screenshot 2023-02-24 at 10 28 20

peetzweg:
Can we reuse these components from the normal market? They are basically always supposed to show the same stuff. If we ever need new column types we can add them. Here a duplication is not beneficial.

Screenshot 2023-02-28 at 10 01 10


Also good to reuse the column from the Pools page. It quite big the new one. Looks like completely different data. The user should be presented with the same element, so he recognizes what data it is easily.


Screenshot 2023-02-28 at 10 07 26

peetzweg:
Let's also make the "chain" the first column. As some tokens are on mutliple chains it's pretty important.

Not sure if we should also move the pool to the front. 🤔 
Let's move the chain only for now.


Screenshot 2023-02-28 at 10 16 43

peetzweg:
Screenshot 2023-02-28 at 15 48 58

peetzweg:
Does it work for you on preview?

I get this error with the dev account. 🤔 

Can we make sure it works even if a single chain might be broken atm? And show a warming toast, that a single chain is currently not reachable?

Screenshot 2023-03-01 at 12 40 40

carlomazzaferro (approved):

PR #1349, opened 2023-03-22
Feat/optimized apr vaults

smartdevcom:


packages/sdk/src/modules/Vaults.ts
```diff
@@ -0,0 +1,225 @@
...
+                  management: managementFee,
+                },
+              ] = await Promise.all([
+                optimizedAPRVault.callStatic.asset(),
+                optimizedAPRVault.callStatic.estimatedTotalAssets(),
+                optimizedAPRVault.callStatic.supplyAPY(0),
+                optimizedAPRVault.callStatic.adapterCount(),
+                optimizedAPRVault.callStatic.emergencyExit(),
+                optimizedAPRVault.callStatic.fees(),
+              ]);
```
vminkov, line 50: these are too many separate RPC calls, let's implement a lens function for these

packages/sdk/src/modules/Vaults.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      return tx;
+    }
+
+    async convertToShares(amount: BigNumber, vault: string) {
+      const optimizedAPRVault = this.createOptimizedAPRVault(vault);
+
+      const totalSupply = await optimizedAPRVault.callStatic.totalSupply();
+      const totalAssets = await optimizedAPRVault.callStatic.totalAssets();
+
```
vminkov, line 139: it looks wrong to me to implement your own `convertToShares` and fee shares calculation.. can you use the `previewDeposit` and `previewWithdraw` fns instead?

packages/sdk/src/modules/Vaults.ts
```diff
@@ -0,0 +1,225 @@
...
+      const { withdrawal: withdrawalFee } = await optimizedAPRVault.callStatic.fees();
+
+      const shares = await this.convertToShares(amount, vault);
+      const feeShares = shares.mul(withdrawalFee).div(utils.parseUnits("1").sub(withdrawalFee));
+
+      // if rounded up, then should add "1"
+      if (!(response.eq(shares.add(feeShares)) || response.eq(shares.add("1").add(feeShares)))) {
+        const errorCode = parseInt(response.toString());
+
+        return { errorCode };
```
vminkov, line 175: the response is never an error code, so this manual calculation and comparing of the returned value is not good

packages/functions/src/controllers/vault-data.ts
```diff
@@ -0,0 +1,90 @@
...
+    const optimizedVaultsRegistry = new ethers.Contract(
+      sdk.chainDeployment.OptimizedVaultsRegistry.address,
+      OptimizedVaultsRegistryABI,
+      sdk.provider
+    ) as OptimizedVaultsRegistry;
+
+    const vaults = await optimizedVaultsRegistry.callStatic.getAllVaults();
+
+    const results = await Promise.all(
+      vaults.map(async (vault) => {
```
vminkov, line 28: the new lens fn can be used here, too

packages/sdk/src/modules/Vaults.ts
smartdevcom: So do I not need to do check with `callStatic` function? and need to directly do `await optimizedAPRVault["deposit(uint256)"](amount);`?

vminkov: confirmed, no need to check with `callStatic`

PR #1383, opened 2023-04-21
save asset price

smartdevcom:
This is implementation of scheduled function to store underlying and usd prices in supabase for all assets in bsc, polygon, moonbeam


packages/functions/src/controllers/asset-price.ts
```diff
@@ -0,0 +1,131 @@
...
+  evmos,
+  fantom,
+  ganache,
+  moonbeam,
+  neondevnet,
+  polygon,
+} from '@midas-capital/chains';
+import axios from 'axios';
+
+export const chainSupportedAssets: ChainSupportedAssets = {
```
carlomazzaferro, line 24: Instead of always defining this, please use: https://github.com/midas-protocol/monorepo/blob/development/packages/chains/src/index.ts#L17

carlomazzaferro (approved):

PR #1408, opened 2023-05-04
Feat/hh tasks fee recipient

vminkov:
## Description

hh task to set the fee recipient for the plugins and flywheels


carlomazzaferro:
Let's add in this PR the setting of the deployer in the AP @vminkov

carlomazzaferro (approved):

PR #1423, opened 2023-05-11
fixed max withdraw for non borrower

smartdevcom:


packages/ui/components/pages/PoolPage/MarketsList/AdditionalInfo/FundButton/WithdrawModal/index.tsx
```diff
@@ -98,7 +98,11 @@ export const WithdrawModal = ({
 
       let resp;
       if (maxWithdrawAmount.eq(amount)) {
-        resp = await currentSdk.withdraw(asset.cToken, constants.MaxUint256);
+        if (asset.borrowBalance.isZero()) {
```
vminkov, line 101: the user borrows are very probably in the other markets in the pool... the easiest way to check if the redeem call will pass is to check with `comptroller.redeemAllowed(marketAddress, userAddress, amount)`

PR #1458, opened 2023-06-12
changed fn

smartdevcom:


packages/sdk/src/modules/Flywheel.ts
```diff
@@ -121,7 +121,8 @@ export function withFlywheel<TBase extends CreateContractsModule = CreateContrac
       for (const flywheel of flywheels) {
```
vminkov, line 121: you don't need to have 2 nested for loops, just call `getClaimRewardsForMarkets` with all `marketsOfPool` and all `flywheels` as input params - the result is return for each reward token of the flywheel in the input

packages/sdk/src/modules/Flywheel.ts
```diff
@@ -111,74 +108,111 @@ export function withFlywheel<TBase extends CreateContractsModule = CreateContrac
...
+      const [, comptrollers] = await fplSecondary.callStatic.getFlywheelsToClaim(account, {
+        from: account,
+      });
+
+      // get all markets of all pools
+      const allMarkets: string[] = [];
+      const allFlywheels: string[] = [];
+
+      await Promise.all(
+        comptrollers.map(async (comptroller) => {
```
vminkov, line 122: let's break down this call, so it is made separately for each pool - so loop for each of the `comptrollers` and call the `this.getClaimableRewardsForMarkets` with `pool.callStatic.getAllMarkets()` for the markets and `pool.callStatic.getRewardsDistributors()` for the flywheels

PR #1460, opened 2023-06-13
fixed estimated usd

smartdevcom:


packages/sdk/src/modules/FusePools.ts
```diff
@@ -96,6 +96,11 @@ export function withFusePools<TBase extends MidasBaseConstructor>(Base: TBase) {
           asset.originalSymbol = _asset.originalSymbol ? _asset.originalSymbol : undefined;
         }
 
+        asset.netSupplyBalance = asset.supplyBalance.sub(asset.borrowBalance);
```
vminkov, line 99: `asset.borrowBalance` can be > than `asset.supplyBalance` but `asset.netSupplyBalance` cannot be negative

if `asset.borrowBalance > asset.supplyBalance` then `asset.netSupplyBalance = 0`

smartdevcom: fixed

vminkov (approved):

PR #1478, opened 2023-06-29
wrap before minting liquidators registry fns

vminkov:
## Description

wrap before mint liquidators registry fns

```
We want to implement in the UI the ability of the user to swap some asset for the underlying of the market before minting. This is an additional step (actually two steps) before the Approve step.

The same functionality can be used also for the Levered Positions, with the difference that it is not an additional step before the Approve step.
```


packages/ui/hooks/useSwapAmount.ts
```diff
@@ -0,0 +1,71 @@
...
+          );
+
+          return { outputAmount, slippage };
+        } catch (e) {
+          console.error(
+            'Could not get swap amount and slippage',
+            { amount, chainId: currentSdk.chainId, inputToken, outputToken },
+            JSON.stringify(e)
+          );
+
```
vminkov, line 63: Can you please change

```
  console.error(
            'Could not get swap amount and slippage',
            { amount, chainId: currentSdk.chainId, inputToken, outputToken },
            JSON.stringify(e)
          );
```

to 


```
  console.error(
            'Could not get swap amount and slippage',
            { amount, chainId: currentSdk.chainId, inputToken, outputToken },
            e
          );
          console.error(
            JSON.stringify(e)
          );

```

packages/sdk/chainDeploy/helpers/liquidators/registry.ts
```diff
@@ -10,22 +10,33 @@ export const configureLiquidatorsRegistry = async ({
...
-    strategies.push(redemptionStrategy.address);
-    inputTokens.push(inputToken);
-    outputTokens.push(outputToken);
+
+    const existingStrategy = strategyMapping[inputToken][outputToken] ?? null;
+    if (existingStrategy != redemptionStrategy.address || existingStrategy == null) {
+      strategies.push(redemptionStrategy.address);
+      inputTokens.push(inputToken);
+      outputTokens.push(outputToken);
+    }
```
vminkov: this will result in configuring only the strategies that are different, but all that are not different will be removed from the on-chain config...

